PR #1035, opened 2015-01-12
Fixed crash in OperationQueue

foxinushka:
There is possible a situation when dispatch_release(mDispatchQueue) in destructor destroys the queue while execution stoppedOnMainThread method.

Test code:

```obj-c
    MCOIMAPSession *session = ...
    
    @autoreleasepool {
        dispatch_queue_t queue = dispatch_queue_create("Test", DISPATCH_QUEUE_SERIAL);
        session.dispatchQueue = queue;
    }
    
    dispatch_async(session.dispatchQueue, ^{
        [[session checkAccountOperation] start:^(NSError *error) {
            completion(error);
            
            dispatch_async(session.dispatchQueue, ^{
                [[session disconnectOperation] start:^(NSError *error) {
                }];
            });
        }];
    });
```

### DIFF
--- a/src/core/basetypes/MCOperationQueue.cpp
+++ b/src/core/basetypes/MCOperationQueue.cpp
@@ -104,7 +104,10 @@ void OperationQueue::runOperations()
             
             retain(); // (2)
 #if __APPLE__
-            performMethodOnDispatchQueue((Object::Method) &OperationQueue::stoppedOnMainThread, NULL, mDispatchQueue, true);
+            dispatch_queue_t queue = mDispatchQueue;
+            dispatch_retain(queue);
+            performMethodOnDispatchQueue((Object::Method) &OperationQueue::stoppedOnMainThread, NULL, queue, true);
+            dispatch_release(queue);
 #else
             performMethodOnMainThread((Object::Method) &OperationQueue::stoppedOnMainThread, NULL, true);
 #endif


dinhviethoa:
I think this fix won't help.
Could you give more details about lifetime of session and of the queue?
Could you also share some crash reports?
A complete file or project that would reproduce the issue would also help.

foxinushka:
In real project I use mail syncer which stores session and queue in strong property. Each mail syncer contains own queue. When I remove syncer, it releases queue & session. If queue executes stoppedOnMainThread while removing it calls "release(); // (3)" and destroy it self and queue where stoppedOnMainThread is executed. It causes to crash. My test code reproduces this situation.
The fix helps because of it retains queue while execution of stoppedOnMainThread.

foxinushka:
The crash:

crash

dinhviethoa:
I fixed it using:
https://github.com/MailCore/mailcore2/commit/dfd76892dbab3b7fdc8dd6e12a2f262b1445a3bc

foxinushka:
Now it crashes in:
screen1

I thought about fix that you done. But it doesn't fix the crash. We should retain the queue before execution a block, then we should release it after execution, but in other queue. If we destroy it while execution it crashes.

dinhviethoa:
http://opensource.apple.com/source/libdispatch/libdispatch-187.5/dispatch/queue.h

oh:
```
 * Unlike dispatch_async(), no retain is performed on the target queue. Because
 * calls to this function are synchronous, the dispatch_sync() "borrows" the
 * reference of the caller.
```

dinhviethoa:
Could you try with that fix? 60f825b5cc92bbfffe1f27f960b9f9d7d226547e

foxinushka:
It works. Thanks.

Issue #236, opened 2013-07-24
Implement validation of email account


dinhviethoa:
**input**

- email / password

or

- email / password
- list of imap (or pop) account
- list of smtp accounts

**output**

- imap server (or pop server)
- login
- password

- smtp server
- login password

DrChrispoper:
Hi, I would like to help with this are there any specifications on how to go with it?
rlaferia was suggesting adding to MCOIMAPSession but if it is to test SMTP too it might not be the best way.
I was thinking of going with something new like MCOAccountValidator and having MCOMailProvider, MCOIMAPSession and MCOSMTPSession as properties of that class. 
Then using like this:

```
self.accountVal = [[MCOAccountValidator alloc]init];
self.accountVal.email = username;
self.accountVal.password = password;
    
[self.accountVal imapStart:^(NSError *error) {
    if (error == nil) {
        //IMAP is ok
    } 
}];

[self.accountVal smtpStart:^(NSError *error) {
    if (error == nil) {
        //SMTP is ok
    }
}];
 ```
With imapStart and smtpStart using checkOperations from MCOIMAP/SMTPOperation.

**Comment by dinhviethoa** [comment]
```
@interface  MCOAccountValidator

@property (nonatomic, copy) NSString * email; /* for SMTP */
@property (nonatomic, copy) NSString * username;
@property (nonatomic, copy) NSString * password;
@property (nonatomic, copy) NSString * OAuth2Token;

@property (nonatomic, retain) NSArray * /* MCONetService */ imapServers;
@property (nonatomic, retain) NSArray * /* MCONetService */ popServers;
@property (nonatomic, retain) NSArray * /* MCONetService */ smtpServers;

// result
@property (nonatomic, retain, readonly) MCONetService * imapServer;
@property (nonatomic, retain, readonly) MCONetService * popServer;
@property (nonatomic, retain, readonly) MCONetService * smtpServer;
@property (nonatomic, retain, readonly) NSError * imapError;
@property (nonatomic, retain, readonly) NSError * popError;
@property (nonatomic, retain, readonly) NSError * smtpError;

- (void) start:(void (^)(void))completionBlock;

@end
```

This class will create the session it needs.
the input array might be nil if it doesn't need to be checked.
The MCOAccountValidator will test everything given as input.

DrChrispoper:
I have a question
There can be different MX records for one hostname but could there be different providers for those MX records?
[...]

PR #1046, opened 2015-01-26
AccountValidator

DrChrispoper:
Implementation of the email and account validation feature #236

### DIFF
--- a/src/cmake/public-headers.cmake
+++ b/src/cmake/public-headers.cmake
@@ -74,6 +74,7 @@ core/provider/MCMailProvidersManager.h
 core/provider/MCMailProvider.h
 core/provider/MCNetService.h
 core/provider/MCAccountValidator.h
+core/provider/MCMXRecordResolverOperation.h
 async/MCAsync.h
 async/smtp/MCAsyncSMTP.h
 async/smtp/MCSMTPAsyncSession.h
@@ -116,7 +117,6 @@ async/nntp/MCNNTPListNewsgroupsOperation.h
 async/nntp/MCNNTPFetchOverviewOperation.h
 async/nntp/MCNNTPFetchServerTimeOperation.h
 async/nntp/MCNNTPOperationCallback.h
-async/mx/MCResolveProviderUsingMXRecord.h
 objc/MCObjC.h
 objc/utils/MCOUtils.h
 objc/utils/MCOObjectWrapper.h
--- a/src/cmake/public-headers.cmake
+++ b/src/cmake/public-headers.cmake
@@ -73,6 +73,7 @@ core/provider/MCProvider.h
 core/provider/MCMailProvidersManager.h
 core/provider/MCMailProvider.h
 core/provider/MCNetService.h
+core/provider/MCAccountValidator.h
 async/MCAsync.h
 async/smtp/MCAsyncSMTP.h
 async/smtp/MCSMTPAsyncSession.h
@@ -115,6 +116,7 @@ async/nntp/MCNNTPListNewsgroupsOperation.h
 async/nntp/MCNNTPFetchOverviewOperation.h
 async/nntp/MCNNTPFetchServerTimeOperation.h
 async/nntp/MCNNTPOperationCallback.h
+async/mx/MCResolveProviderUsingMXRecord.h
 objc/MCObjC.h
 objc/utils/MCOUtils.h
 objc/utils/MCOObjectWrapper.h
@@ -200,4 +202,5 @@ objc/provider/MCOProvider.h
 objc/provider/MCONetService.h
 objc/provider/MCOMailProvider.h
 objc/provider/MCOMailProvidersManager.h
+objc/provider/MCOAccountValidator.h
 )
--- a/src/cmake/public-headers.cmake
+++ b/src/cmake/public-headers.cmake
@@ -74,7 +74,6 @@ core/provider/MCMailProvidersManager.h
 core/provider/MCMailProvider.h
 core/provider/MCNetService.h
 core/provider/MCAccountValidator.h
-core/provider/MCMXRecordResolverOperation.h
 async/MCAsync.h
 async/smtp/MCAsyncSMTP.h
 async/smtp/MCSMTPAsyncSession.h
--- a/src/core/provider/MCAccountValidator.cpp
+++ b/src/core/provider/MCAccountValidator.cpp
@@ -228,6 +228,7 @@ void AccountValidator::checkNextHost()
             mImapSession->setConnectionType(mImapServer->connectionType());
             
             mOperation = (IMAPOperation *)mImapSession->checkAccountOperation();
+            mOperation->retain();
             mOperation->setCallback(this);
             mOperation->start();
         }
@@ -249,6 +250,7 @@ void AccountValidator::checkNextHost()
             mPopSession->setConnectionType(mPopServer->connectionType());
             
             mOperation = mPopSession->checkAccountOperation();
+            mOperation->retain();
             mOperation->setCallback(this);
             mOperation->start();
         }
@@ -270,6 +272,7 @@ void AccountValidator::checkNextHost()
             mSmtpSession->setConnectionType(mSmtpServer->connectionType());
             
             mOperation = mSmtpSession->checkAccountOperation(Address::addressWithMailbox(mEmail));
+            mOperation->retain();
             mOperation->setCallback(this);
             mOperation->start();
         }
@@ -303,6 +306,8 @@ void AccountValidator::checkNextHostDone()
         error = mSmtpError;
         MC_SAFE_RELEASE(mSmtpSession);
     }
+    
+    MC_SAFE_RELEASE(mOperation);
     
     if (error == ErrorNone) {
         mCurrentServiceTested ++;
--- a/src/core/provider/MCAccountValidator.cpp
+++ b/src/core/provider/MCAccountValidator.cpp
@@ -17,9 +17,8 @@
 #include "MCIMAPOperation.h"
 #include "MCPOPOperation.h"
 #include "MCSMTPOperation.h"
-
-#include <arpa/inet.h>
-#include <resolv.h>
+#include "MCResolveProviderUsingMXRecord.h"
+#include "MCOperationQueueCallback.h"
 
 using namespace mailcore;
 
@@ -30,6 +29,35 @@ enum {
     SERVICE_POP,   /* POP  service */
     SERVICE_SMTP,  /* SMTP service */
 };
+
+namespace mailcore {
+    
+    class ValidatorOperationQueueCallback : public Object, public OperationQueueCallback {
+    public:
+        ValidatorOperationQueueCallback(AccountValidator * validator) {
+            mValidator = validator;
+        }
+        
+        virtual ~ValidatorOperationQueueCallback() {
+        }
+        
+        virtual void queueStartRunning() {
+            if (mValidator->operationQueueCallback() != NULL) {
+                mValidator->operationQueueCallback()->queueStartRunning();
+            }
+        }
+        
+        virtual void queueStoppedRunning() {
+            if (mValidator->operationQueueCallback() != NULL) {
+                mValidator->operationQueueCallback()->queueStoppedRunning();
+            }
+    
+            mValidator->setup();
+        }
+    private:
+        AccountValidator * mValidator;
+    };
+}
 
 void AccountValidator::init()
 {
@@ -52,6 +80,13 @@ void AccountValidator::init()
     
     mCurrentServiceIndex = 0;
     mCurrentServiceTested = 0;
+    
+    mProvider = NULL;
+    
+    mQueue = new OperationQueue();
+    mQueueCallback = new ValidatorOperationQueueCallback(this);
+    mQueue->setCallback(mQueueCallback);
+    mOperationQueueCallback = NULL;
 }
 
 AccountValidator::AccountValidator()
@@ -69,12 +104,46 @@ AccountValidator::~AccountValidator()
     MC_SAFE_RELEASE(mSmtpServices);
     MC_SAFE_RELEASE(mPopServices);
     MC_SAFE_RELEASE(mIdentifier);
+    MC_SAFE_RELEASE(mProvider);
 }
 
 void AccountValidator::start()
 {
-    setup();
-    test();
+    if (mEmail == NULL) {
+        if (mUsername == NULL) {
+            return;
+        }
+        else {
+            mEmail = mUsername;
+        }
+    }
+    else if (mUsername == NULL){
+        mUsername = mEmail;
+    }
+    
+    mProvider = MailProvidersManager::sharedManager()->providerForEmail(mUsername);
+    
+    if (mProvider == NULL) {
+        Array * components;
+        String * domain;
+        
+        components = mUsername->componentsSeparatedByString(MCSTR("@"));
+        if (components->count() >= 2) {
+            domain = (String *) components->lastObject();
+            mResolveMX = new ResolveProviderUsingMXRecord();
+            mResolveMX->setHostname(domain);
+            mQueue->addOperation(mResolveMX);
+        }
+    }
+    else{
+        setup();
+    }
+}
+
+void AccountValidator::cancel()
+{
+    mOperation->cancel();
+    Operation::cancel();
 }
 
 void AccountValidator::operationFinished(Operation * op)
@@ -84,38 +153,21 @@ void AccountValidator::operationFinished(Operation * op)
 
 void AccountValidator::setup()
 {
-    if (mEmail == NULL) {
-        if (mUsername == NULL) return;
-        else mEmail = mUsername;
-    }else if (mUsername == NULL){
-        mUsername = mEmail;
+    if (mResolveMX != NULL) {
+       mProvider = mResolveMX->provider();
     }
     
-    MailProvider *provider = MailProvidersManager::sharedManager()->providerForEmail(mUsername);
-    
-    if (provider == NULL) {
-        Array * components;
-        String * domain;
-    
-        components = mUsername->componentsSeparatedByString(MCSTR("@"));
-        if (components->count() < 2)
-            return;
-    
-        domain = (String *) components->lastObject();
-        provider = ResolveProviderUsingMXRecord(domain);
-    }
-    
-    if (provider != NULL) {
-        mIdentifier = provider->identifier();
+    if (mProvider != NULL) {
+        mIdentifier = mProvider->identifier();
         
-        if (mImapServices->count() == 0 and provider->imapServices()->count() > 0)
-            mImapServices = provider->imapServices();
+        if (mImapServices->count() == 0 and mProvider->imapServices()->count() > 0)
+            mImapServices = mProvider->imapServices();
         
-        if (mPopServices->count() == 0 and provider->popServices()->count() > 0)
-            mPopServices = provider->popServices();
+        if (mPopServices->count() == 0 and mProvider->popServices()->count() > 0)
+            mPopServices = mProvider->popServices();
         
-        if (mSmtpServices->count() == 0 and provider->smtpServices()->count() > 0)
-            mSmtpServices = provider->smtpServices();
+        if (mSmtpServices->count() == 0 and mProvider->smtpServices()->count() > 0)
+            mSmtpServices = mProvider->smtpServices();
     }
     
     if (mImapServices->count() == 0)
@@ -126,6 +178,8 @@ void AccountValidator::setup()
     
     if (mSmtpServices->count() == 0)
         mSmtpError = ErrorNoValidServerFound;
+    
+    test();
 }
 
 void AccountValidator::opCompleted()
@@ -133,16 +187,22 @@ void AccountValidator::opCompleted()
     ErrorCode error = ErrorNone;
     
     if (mCurrentServiceTested == SERVICE_IMAP) {
-        error = mImapError = ((IMAPOperation *)mOperation)->error();
-    } else if (mCurrentServiceTested == SERVICE_POP) {
-        error = mPopError = ((POPOperation *)mOperation)->error();
-    } else if (mCurrentServiceTested == SERVICE_SMTP) {
-        error = mSmtpError = ((SMTPOperation *)mOperation)->error();
+        mImapError = ((IMAPOperation *)mOperation)->error();
+        error = mImapError;
+    }
+    else if (mCurrentServiceTested == SERVICE_POP) {
+        mPopError = ((POPOperation *)mOperation)->error();
+        error = mPopError;
+    }
+    else if (mCurrentServiceTested == SERVICE_SMTP) {
+        mSmtpError = ((SMTPOperation *)mOperation)->error();
+        error = mSmtpError;
     }
     
     if (error == ErrorNone) {
         mCurrentServiceTested ++;
-    } else {
+    }
+    else {
         mCurrentServiceIndex ++;
     }
     
@@ -173,12 +233,14 @@ void AccountValidator::test()
             mOperation->setCallback(this);
             mOperation->start();
         
-        } else {
+        }
+        else {
             mCurrentServiceTested = SERVICE_POP;
             mCurrentServiceIndex = 0;
             test();
         }
-    }else if (mCurrentServiceTested == SERVICE_POP){
+    }
+    else if (mCurrentServiceTested == SERVICE_POP){
         if (mCurrentServiceIndex < mPopServices->count()) {
             POPAsyncSession *popSession = new POPAsyncSession();
             popSession->setUsername(mUsername);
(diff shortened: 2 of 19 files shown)


src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,384 @@
...
+            }
+    }
+    }
+    
+    return (mImapServices->count() > 0 || mPopServices->count() > 0 || mSmtpServices->count() > 0);
+}
+
+void AccountValidator::opCompleted()
+{
+    if (mCurrentServiceTested == 0) {
```
dinhviethoa: use constants such as SERVICE_IMAP, SERVICE_POP, SERVICE_SMTP.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,384 @@
...
+        mImapError = ((IMAPOperation *)mOperation)->error();
+        (mImapError == ErrorNone) ? mCurrentServiceTested ++ : mCurrentServiceIndex ++;
```
dinhviethoa: Use if / else with brace instead.
You could also refactor some code: error = mImapError;
if (error == ErrorNone) {
  // next type of service.
  mCurrentServiceTested ++;
}
else {
  // test next server.
  mCurrentServiceIndex ++;
}

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+        } else {
+           mCallback->operationFinished(this);
+        }
+    } else {
+        mCallback->operationFinished(this);
+    }
+}
+
+MailProvider* AccountValidator::VTMxRecordForHostname(String *hostname)
```
dinhviethoa: Suggestion: ResolveProviderUsingMXRecord(String * domain)

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,384 @@
...
+    MC_SAFE_RELEASE(mOAuth2Token);
+    MC_SAFE_RELEASE(mImapServices);
+    MC_SAFE_RELEASE(mSmtpServices);
+    MC_SAFE_RELEASE(mPopServices);
+    MC_SAFE_RELEASE(mIdentifier);
+}
+
+void AccountValidator::start()
+{
+    if (setup()) test();
```
dinhviethoa: The following should just work:
```
{
  setup();
  test();
}
```

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,384 @@
...
+        if (mSmtpServices->count() == 0) {
+            if (provider->smtpServices()->count() > 0) {
+                mSmtpServices = provider->smtpServices();
+            } else {
+                mSmtpError = ErrorNoValidServerFound;
+            }
+    }
+    }
+    
+    return (mImapServices->count() > 0 || mPopServices->count() > 0 || mSmtpServices->count() > 0);
```
dinhviethoa: You probably don't need to return any value.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,384 @@
...
+            mOperation = (IMAPOperation *)imapSession->checkAccountOperation();
+            mOperation->setCallback(this);
+            mOperation->start();
+        
+        } else {
+            mCurrentServiceTested = 1;
+            mCurrentServiceIndex = 0;
+            test();
+        }
+    }else if (mCurrentServiceTested == 1){
```
dinhviethoa: style:
```
}
else if () {
```

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,384 @@
...
+    if (provider == NULL) {
+        Array * components;
+        String * domain;
+    
+        components = mUsername->componentsSeparatedByString(MCSTR("@"));
+        if (components->count() < 2)
+            return false;
+    
+        domain = (String *) components->lastObject();
+        provider = VTMxRecordForHostname(domain);
```
dinhviethoa: VTMxRecordForHostname() should probably not run in the main thread because it's performing a network call. I'd suggest creating an operation and adding an OperationQueue for that.

src/core/provider/MCAccountValidator.h
```diff
@@ -0,0 +1,96 @@
...
+        // result
+        virtual String * identifier();
+        virtual NetService * imapServer();
+        virtual NetService * popServer();
+        virtual NetService * smtpServer();
+        virtual ErrorCode imapError();
+        virtual ErrorCode popError();
+        virtual ErrorCode smtpError();
+        
+        virtual void start();
```
dinhviethoa: I think we need a cancel() method here.

src/core/provider/MCAccountValidator.h
```diff
@@ -0,0 +1,96 @@
...
+        
+        virtual void setCallback(OperationCallback * callback);
```
dinhviethoa: You don't need them. They're already defined in Operation.

src/core/provider/MCAccountValidator.h
DrChrispoper: Isn't the cancel() from Operation called? 
In which case I should still add something like 
if (isCancelled())
        return;
At the start of my operationFinished method?

dinhviethoa: You have to override cancel() and cancel all the pending operations.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,387 @@
...
+}
+
+void AccountValidator::operationFinished(Operation * op)
+{
+    opCompleted();
+}
+
+void AccountValidator::setup()
+{
+    if (mEmail == NULL) {
```
dinhviethoa:
```
if (mEmail == NULL) {
  if (mUsername == NULL) {
    return;
   }
   else {
     mEmail = mUsername;
   }
}
else if (mUsername == NULL) {
  mUsername = mEmail;
}
```

dinhviethoa: There are other locations where if else should follow that guide.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,387 @@
...
+    if (mSmtpServices->count() == 0)
+        mSmtpError = ErrorNoValidServerFound;
+}
+
+void AccountValidator::opCompleted()
+{
+    ErrorCode error = ErrorNone;
+    
+    if (mCurrentServiceTested == SERVICE_IMAP) {
+        error = mImapError = ((IMAPOperation *)mOperation)->error();
```
dinhviethoa: two lines:

```
mImapError = ((IMAPOperation *)mOperation)->error();
error = mImapError;
```

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,387 @@
...
+    if (provider == NULL) {
+        Array * components;
+        String * domain;
+    
+        components = mUsername->componentsSeparatedByString(MCSTR("@"));
+        if (components->count() < 2)
+            return;
+    
+        domain = (String *) components->lastObject();
+        provider = ResolveProviderUsingMXRecord(domain);
```
dinhviethoa: You still need to run ResolveProviderUsingMXRecord() asynchronously.
We should not block the main thread while running it.

src/core/provider/MCAccountValidator.cpp
DrChrispoper: Ok. 
Wasn't sure if it was just the else if() you wanted the line return.

src/core/provider/MCAccountValidator.cpp
DrChrispoper: Sorry didn't properly refresh the page so didn't see all the comments at first.
Will do that.

DrChrispoper: Do I need to add -lresolv to mailcore linker flags?

dinhviethoa:
You also need to add the additional header to public-headers.cmake to get the build on Linux to succeed. Linux probably doesn't need -lresolv. It's probably needed on the Mac. Don't forget to add it on the CMake build too.

DrChrispoper:
Is there a reason why the provider files are not in the objc.cmake?

dinhviethoa:
I think I missed that. You can add them.

src/core/provider/MCAccountValidator.h
```diff
@@ -0,0 +1,109 @@
...
+        virtual NetService * popServer();
+        virtual NetService * smtpServer();
+        virtual ErrorCode imapError();
+        virtual ErrorCode popError();
+        virtual ErrorCode smtpError();
+        
+        virtual void start();
+        virtual void cancel();
+        
+        virtual void setOperationQueueCallback(OperationQueueCallback * callback);
```
dinhviethoa: Remove those methods.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,412 @@
...
+{
+    return mPopError;
+}
+
+ErrorCode AccountValidator::smtpError()
+{
+    return mSmtpError;
+}
+
+void POPAsyncSession::setOperationQueueCallback(OperationQueueCallback * callback)
```
dinhviethoa: You probably meant AccountValidator:: setOperationQueueCallback().
Anyway remove those methods.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,412 @@
...
+            imapSession->setPort(mImapServer->port());
+            imapSession->setConnectionType(mImapServer->connectionType());
+            
+            mOperation = (IMAPOperation *)imapSession->checkAccountOperation();
+            mOperation->setCallback(this);
+            mOperation->start();
+        
+        }
+        else {
+            mCurrentServiceTested = SERVICE_POP;
```
dinhviethoa: mCurrentServiceTested ++;

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,412 @@
...
+            smtpSession->setPort(mSmtpServer->port());
+            smtpSession->setConnectionType(mSmtpServer->connectionType());
+            
+            mOperation =  (SMTPOperation *)smtpSession->checkAccountOperation(Address::addressWithMailbox(mEmail));
+            mOperation->setCallback(this);
+            mOperation->start();
+
+        }
+        else {
+           callback()->operationFinished(this);
```
dinhviethoa: Which makes this code consistent with the other ones:
mCurrentServiceTested ++;
mCurrentServiceIndex = 0;
test();

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,412 @@
...
+    
+    if (mSmtpServices->count() == 0)
+        mSmtpError = ErrorNoValidServerFound;
+    
+    test();
+}
+
+void AccountValidator::opCompleted()
+{
+    ErrorCode error = ErrorNone;
```
dinhviethoa: You'll have to check if the operation is mResolveMX.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/**
+ Each service(IMAP/POP/SMTP) is tested one after the other.
+ For each service we test each server details (NetService),
+ Until either:  
+    we find on that works and returns ErrorNone in opCompleted().
+    we have gone trough the Array of NetService for that service and test() is recalled for the next service.
+ */
+void AccountValidator::test()
```
dinhviethoa: rename to AccountValidator::checkNextHost()

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+void AccountValidator::cancel()
+{
+    mOperation->cancel();
+    Operation::cancel();
+}
+
+void AccountValidator::operationFinished(Operation * op)
+{
+    opCompleted();
```
dinhviethoa: The body of this method should be:
```
if (op == mResolveMX) {
  resolveMXDone();
}
else {
  checkNextHostDone();
}
```

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,412 @@
...
+ For each service we test each server details (NetService),
+ Until either:  
+    we find on that works and returns ErrorNone in opCompleted().
+    we have gone trough the Array of NetService for that service and test() is recalled for the next service.
+ */
+void AccountValidator::test()
+{
+    if (mCurrentServiceTested == SERVICE_IMAP) {
+        if (mCurrentServiceIndex < mImapServices->count()) {
+            IMAPAsyncSession *imapSession = new IMAPAsyncSession();
```
dinhviethoa: It would be nice to keep the IMAP session around.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,412 @@
...
+    if (mPopServices->count() == 0)
+        mPopError = ErrorNoValidServerFound;
+    
+    if (mSmtpServices->count() == 0)
+        mSmtpError = ErrorNoValidServerFound;
+    
+    test();
+}
+
+void AccountValidator::opCompleted()
```
dinhviethoa: move it below checkNextHost() and rename it to checkNextHostDone().

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,412 @@
...
+    MC_SAFE_RELEASE(mOAuth2Token);
+    MC_SAFE_RELEASE(mImapServices);
+    MC_SAFE_RELEASE(mSmtpServices);
+    MC_SAFE_RELEASE(mPopServices);
+    MC_SAFE_RELEASE(mIdentifier);
+    MC_SAFE_RELEASE(mProvider);
+}
+
+void AccountValidator::start()
+{
```
dinhviethoa: void AccountValidator::start()
{
  ...
  mProvider = MailProvidersManager::sharedManager()->providerForEmail(mUsername);
  if (mProvider == NULL) {
    resolveMX();
  }
  else {
    checkHosts();
  }
}

void AccountValidator::resolveMX()
{
...
}

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,412 @@
...
+    mOperation->cancel();
+    Operation::cancel();
+}
+
+void AccountValidator::operationFinished(Operation * op)
+{
+    opCompleted();
+}
+
+void AccountValidator::setup()
```
dinhviethoa: rename to startCheckingHosts()

dinhviethoa:
Lots of comments but we're making good progress in the right direction.

DrChrispoper:
Thanks for all the feedback!

src/core/provider/MCAccountValidator.cpp
DrChrispoper: I've added the sessions as properties so we can keep the IMAP session and properly release the sessions others at the end of the operation. Or release them all if canceled.
Do you want to keep the session to maybe make it accessible after the operation?

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,421 @@
...
+        mQueue = new OperationQueue();
+        mQueue->addOperation(mResolveMX);
+    }
+}
+
+void AccountValidator::resolveMXDone()
+{
+    Array * mxRecords = mResolveMX->mxRecords();
+    
+    CFURLRef imageURL = CFBundleCopyResourceURL(CFBundleGetMainBundle(), CFSTR("providers"), CFSTR("json"), NULL);
```
dinhviethoa: Could you just suppose that MailProvidersManager::sharedManager() is already pre-filled? and remove all these lines 162-167.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,421 @@
...
+    }
+}
+
+void AccountValidator::resolveMX()
+{
+    Array * components;
+    String * domain;
+    
+    components = mUsername->componentsSeparatedByString(MCSTR("@"));
+    if (components->count() >= 2) {
```
dinhviethoa: what happens if components->count() < 2? You should probably fail.

src/core/provider/MCMXRecordResolverOperation.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+MXRecordResolverOperation::MXRecordResolverOperation()
+{
+    mHostname = NULL;
+    mMXRecords = NULL;
+}
+
+MXRecordResolverOperation::~MXRecordResolverOperation()
+{
+     MC_SAFE_RELEASE(mMXRecords);
```
dinhviethoa: release mHostname too.

src/cmake/public-headers.cmake
```diff
@@ -73,6 +73,8 @@ core/provider/MCProvider.h
 core/provider/MCMailProvidersManager.h
 core/provider/MCMailProvider.h
 core/provider/MCNetService.h
+core/provider/MCAccountValidator.h
+core/provider/MCMXRecordResolverOperation.h
```
dinhviethoa: I don't think MCMXRecordResolverOperation.h needs to be public. Remove it.

dinhviethoa:
Sorry about the delay. Try to ping me if you have no news.
I added some small comments.

DrChrispoper:
It's ok. I don't expect you to be working on this 24/7 :)

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,421 @@
...
+        if (mCurrentServiceIndex < mImapServices->count()) {
+            
+            if (mImapSession == NULL) {
+                mImapSession = new IMAPAsyncSession();
+                mImapSession->setUsername(mUsername);
+                mImapSession->setPassword(mPassword);
+            }
+            
+            mImapServer = (NetService *) mImapServices->objectAtIndex(mCurrentServiceIndex);
+            mImapSession->setHostname(mImapServer->hostname());
```
dinhviethoa: Create a new session each time you test. Sessions are not designed to be reusable with different servers.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,421 @@
...
+    }
+    else{
+        startCheckingHosts();
+    }
+}
+
+void AccountValidator::cancel()
+{
+    if(mOperation != NULL)
+        mOperation->cancel();
```
dinhviethoa: MC_SAFE_RELEASE(mOperation);

and same below.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,421 @@
...
+
+        }
+        else {
+            mCurrentServiceTested ++;
+            mCurrentServiceIndex = 0;
+            checkNextHost();
+        }
+    }
+    else {
+        MC_SAFE_RELEASE(mPopSession);
```
dinhviethoa: MC_SAFE_RELEASE(mImapSession) too.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,421 @@
...
+    }
+    else if (mCurrentServiceTested == SERVICE_POP) {
+        mPopError = ((POPOperation *)mOperation)->error();
+        error = mPopError;
+    }
+    else if (mCurrentServiceTested == SERVICE_SMTP) {
+        mSmtpError = ((SMTPOperation *)mOperation)->error();
+        error = mSmtpError;
+    }
+
```
dinhviethoa: You should probably release the sessions and the mOperation here.

src/core/provider/MCMXRecordResolverOperation.cpp
```diff
@@ -0,0 +1,73 @@
...
+    unsigned char response[NS_PACKETSZ];
+    ns_msg handle;
+    ns_rr rr;
+    int len;
+    char dispbuf[4096];
+    
+    if (((len = res_search(MCUTF8(mHostname), ns_c_in, ns_t_mx, response, sizeof(response))) >= 0) and
+        (ns_initparse(response, len, &handle) >= 0) and
+        (ns_msg_count(handle, ns_s_an) >= 0)) {
+
```
dinhviethoa: remove blank lines.

src/objc/provider/MCOAccountValidator.mm
```diff
@@ -0,0 +1,154 @@
...
+}
+
+MCO_OBJC_SYNTHESIZE_STRING(setEmail, email)
+MCO_OBJC_SYNTHESIZE_STRING(setUsername, username)
+MCO_OBJC_SYNTHESIZE_STRING(setPassword, password)
+MCO_OBJC_SYNTHESIZE_STRING(setOAuth2Token, OAuth2Token)
+
+- (id) init
+{
+    mailcore::AccountValidator * msg = new mailcore::AccountValidator();
```
dinhviethoa: rename msg to validator.

src/core/provider/MCAccountValidator.cpp
DrChrispoper: I had purposely left it out because you mentioned earlier about keeping the IMAPSession because it might be useful. Should I just not think about that for now?

src/core/provider/MCAccountValidator.cpp
DrChrispoper: If I release each session here after each use. 
This means I do not need to do it at the of checkNextHost() just before callback()->operationFinished(this); right?

dinhviethoa: That's correct.  

--  
Hoa V. Dinh

src/core/provider/MCAccountValidator.cpp
dinhviethoa: I was just advising to keep the session in an ivar around while the operation is still running.
When the operation is done. It's ok to get release of the session.

DrChrispoper: Ok. Makes sense.



—
Christopher Hockley

src/core/provider/MCAccountValidator.cpp
dinhviethoa: You should release mOperation and mResolveMX. You should test mQueue against NULL and set mQueue in init().

DrChrispoper: I am releasing mOperation, mResolveMX and mQueue below.
mQueue is tested against NULL below too. Is that what you mean? 
Do you mean initialize mQueue in init() rather than later in resolveMX()?

dinhviethoa:
I think it's in a good enough state and ready to merge.
Have you tried it after the additional couple of commits?

DrChrispoper:
I'm having issues actually with the SMTP both when the password is wrong and it tries several settings (crashes on second or third time) and with two step verification.
I'll try and see what's wrong.

DrChrispoper:
Ok, I seem to have found out the issue.
After checkNextHost() has run to check a setting, I am releasing the session that was used and the mOperation in checkNextHostDone().
Releasing the mOperation seems to mess up with the OperationQueue (it was running in), in the runOperations() when the pool is released.

Should I not release the mOperation then as it is (if I understood right) autoreleased in the Queue?

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,420 @@
...
+            mImapSession = new IMAPAsyncSession();
+            mImapSession->setUsername(mUsername);
+            mImapSession->setPassword(mPassword);
+            
+            mImapServer = (NetService *) mImapServices->objectAtIndex(mCurrentServiceIndex);
+            mImapSession->setHostname(mImapServer->hostname());
+            mImapSession->setPort(mImapServer->port());
+            mImapSession->setConnectionType(mImapServer->connectionType());
+            
+            mOperation = (IMAPOperation *)mImapSession->checkAccountOperation();
```
dinhviethoa: You should retain the operation here.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,420 @@
...
+            mPopSession->setPassword(mPassword);
+            
+            mPopServer = (NetService *) mPopServices->objectAtIndex(mCurrentServiceIndex);
+            mPopSession->setHostname(mPopServer->hostname());
+            mPopSession->setPort(mPopServer->port());
+            mPopSession->setConnectionType(mPopServer->connectionType());
+            
+            mOperation = mPopSession->checkAccountOperation();
+            mOperation->setCallback(this);
+            mOperation->start();
```
dinhviethoa: same here.

src/core/provider/MCAccountValidator.cpp
```diff
@@ -0,0 +1,420 @@
...
+            mSmtpSession->setPassword(mPassword);
+            
+            mSmtpServer = (NetService *) mSmtpServices->objectAtIndex(mCurrentServiceIndex);
+            mSmtpSession->setHostname(mSmtpServer->hostname());
+            mSmtpSession->setPort(mSmtpServer->port());
+            mSmtpSession->setConnectionType(mSmtpServer->connectionType());
+            
+            mOperation = mSmtpSession->checkAccountOperation(Address::addressWithMailbox(mEmail));
+            mOperation->setCallback(this);
+            mOperation->start();
```
dinhviethoa: and here.

dinhviethoa:
I added some comments about how to fix properly the issue.

DrChrispoper:
Thanks!

DrChrispoper:
I've tested the common use cases and it's working properly now.

dinhviethoa:
Do we need to change the doc to reflect the -lresolv flags for the linker?
To which platform does it apply? OS X / iOS or/and Linux?

DrChrispoper:
We need to change the doc for iOS and OS X.
For Linux I haven't made any changes, I have no means to testing unfortunately.
There still is the question of the MailProvidersManager being pre-filled with the providers file.

dinhviethoa:
It sounds like the unit tests pass on linux. It means that there's no need for additional linker flags.

dinhviethoa:
Merged. Thanks a lot for this work!

PR #1048, opened 2015-01-27
Add support for live.co.uk

ajcollins:
Hello,

Whilst live.co.uk is not available for new signups, I've added it to support their current users.

Best wishes,

Alex


dinhviethoa:
Thanks!

PR #1117, opened 2015-05-04
flag change to ensure ios example can compile

frankobe:
1. remove "-luchardet-ios" 
2. add "-lresolv" flag


dinhviethoa:
Thanks!

Issue #1144, opened 2015-06-17
Problem about “Undisclosed recipients:” when save draft mail without To, Cc, Bcc


JoeTotee:
Hi,

I have trouble with saving draft without To, Cc, Bcc to Gmail via mailcore2. MCOMessageBuilder.data automatically adds an empty group, “Undisclosed recipients:”, to To field. When I open this draft on Gmail webmail, it shows invalid “Undisclosed recipients:” in To field.

However, when I try to save empty To, Cc ,Bcc on Gmail webmail, then open on my iOS app, it shows empty To field.

I investigated “Undisclosed recipients:” on Gmail, Outlook and Yahoo and found that when save draft, they do not add “Undisclosed recipients:” empty group into To field of draft mail. Should “Undisclosed recipients:” empty group be automatically added when sending mail only, not when saving draft?

dinhviethoa:
Would you feel comfortable and send a pull request to implement it?
There's already a similar functionality:
dataAndFilterBccAndForEncryption()
"FilterBcc" might also not add the "undisclosed-recipient" group.
And we could rename "FilterBcc" to "ForDraft".

dinhviethoa:
Fixed with https://github.com/MailCore/mailcore2/commit/4b3ee932f60b6dc15116127f93e3471b74876d62

PR #1148, opened 2015-06-19
Filter Undisclosed recipient for draft message

JoeTotee:
https://github.com/MailCore/mailcore2/issues/1144

### DIFF
--- a/src/core/abstract/MCMessageHeader.cpp
+++ b/src/core/abstract/MCMessageHeader.cpp
@@ -678,6 +678,11 @@ static clist * msg_id_from_string_array(Array * msgids)
 
 struct mailimf_fields * MessageHeader::createIMFFieldsAndFilterBcc(bool filterBcc)
 {
+  return createIMFFieldsForDraftAndFilterBcc(false, filterBcc);
+}
+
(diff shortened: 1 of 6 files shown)


dinhviethoa:
Hi @JoeTotee, you patch looks good.
However, I was concerned adding an additional method (-dataForDraft).
Therefore, I implemented a different way to achieve the behavior you want (https://github.com/MailCore/mailcore2/commit/4b3ee932f60b6dc15116127f93e3471b74876d62).

dinhviethoa:
I'm closing that pull request.
Let me know if everything work for you.

JoeTotee:
Thank you for your patch.  
Everything works fine when saving a draft.

However, I still have some troubles. My customers need to send mail without ‘Undisclosed Recipient’.

I investigated about ‘Undisclosed Recipient’, and found that the standard doesn’t precisely define when to add ‘Undisclosed Recipient’.

Furthermore, I also found that Outlook and Yahoo don’t add ‘Undisclosed Recipient’ empty group to To field of their Bcc-only mail but Gmail does.

I think MailCore2 applies Gmail style. It automatically adds ‘Undisclosed Recipient’ to To field of Bcc-only mail, but my customers use Outlook and they prefer Outlook style.

Is there any way to send Bcc mail without ‘Undisclosed Recipient’ empty group via MailCore2?

PR #1149, opened 2015-06-19
Fixed MCOIMAPIdentity initialization bug

finian:
There was no init method to initialize the native type instance variable `_nativeIdentity`. So codes like below will crash:
```objective-c
MCOIMAPIdentity *identity = [MCOIMAPIdentity identityWithVendor:@"Mozilla" name:@"Thunderbird" version:@"17.0.5"];
```


src/objc/imap/MCOIMAPIdentity.mm
```diff
@@ -33,6 +33,11 @@ - (id) initWithMCIdentity:(mailcore::IMAPIdentity *)identity
     return self;
 }
 
+- (id) init
+{
+    return [self initWithMCIdentity:new nativeType()];
```
dinhviethoa: the allocated nativeType() needs to be released.
Take a look at MCOIMAPFolder for reference.

finian:
Created a new PR.

PR #1151, opened 2015-06-22
Fixed MCOIMAPIdentity initialization bug

finian:
There was no init method to initialize the native type instance variable `_nativeIdentity`. So codes like below will crash:
```objective-c
MCOIMAPIdentity *identity = [MCOIMAPIdentity identityWithVendor:@"Mozilla" name:@"Thunderbird" version:@"17.0.5"];
```


dinhviethoa:
Thanks!

Issue #1164, opened 2015-07-06
iPhone: Attaching a Folder does NOT zip it on the device but does on the simulator


juliangsp:
iPhone 6 & 6 Plus - iOS v8.3 and 8.4; obtaining code from Cocoapod:

When trying to attach a directory/folder using:
MCOAttachment *attachFolder = [MCOAttachment attachmentWithContentsOfFile:path];

on the simulator it does compress/zip the folder (everything works fine) but on the actual iPhone device the attachmentWithContentsOfFile: method returns nil, and there for it is not able to compress the folder.

If I attach just a file (not a folder to be compressed) it works fine on both: iPhone and the simulator.

juliangsp:
The problem seems to be that on the iPhone the method "mailcore::CreateTemporaryZipFileFromFolder" (@ MCZip.cpp) is trying to create a directory /tmp/mailcore2-XXXXXX which it's fine on the simulator (OS X) but iOS on the phone does not allow it.

dinhviethoa:
I think we would need to create a MCZipMac.mm, create a function mailcore::String * mailcore::TemporaryDirectoryForZip(); and use NSTemporaryFolder() to implement it.
then, use this function in CreateTemporaryZipFileFromFolder() in the case of the iOS/Mac.

Let me know if you'd like to create a pull request to fix it.

juliangsp:
Yes, sure I'll like to do a pull request for the fix. Can I create a branch instead of forking the whole repo?

dinhviethoa:
I think you have to fork the whole repo per github rulese.

juliangsp:
I forked and uploaded a fix, but for now it only works for the osx and ios targets since it's mixing Objective-C and C++ code. I don't have an idea on how to best have this code be separated for when compiling for other platforms.

What would you do, dinhviethoa?

You can have a look at the code at my forked repo in branch '#1164_issue_fix_iosmac ' at:
https://github.com/juliangsp/mailcore2/tree/%231164_issue_fix_iosmac

dinhviethoa:
Added comments on your commit.

juliangsp:
New commit after your review:
https://github.com/juliangsp/mailcore2/commit/b1b7587fe8acccbfe46fe59186ec6f383f6d91a6

PR #1169, opened 2015-07-09
#1164 issue fix iosmac

juliangsp:
Fix for issue #1164 ; previous code reviews by dinhviethoa.


juliangsp:
__APPLE__ is a predefine macro for OSX and iOS; there's no need to further define it in MCZipMac.h

src/core/zip/MCZipMac.mm
```diff
@@ -0,0 +1,31 @@
...
+{
+    NSError * error;
+    NSString * newDirString;
+    NSURL * directoryURL;
+    
+    error = nil;
+    newDirString = [NSTemporaryDirectory() stringByAppendingPathComponent:[[NSProcessInfo processInfo] globallyUniqueString]];
+    directoryURL = [NSURL fileURLWithPath:newDirString isDirectory:YES];
+    
+    [[NSFileManager defaultManager] createDirectoryAtURL:directoryURL withIntermediateDirectories:YES attributes:nil error:&error];
```
dinhviethoa: You could use `- (BOOL)createDirectoryAtPath:(NSString * nonnull)path withIntermediateDirectories:(BOOL)createIntermediates attributes:(NSDictionary<NSString *,id> * nullable)attributes error:(NSError * nullable * nullable)error` instead.

src/core/zip/MCZipMac.mm
```diff
@@ -0,0 +1,31 @@
...
+    
+    if (error != nil)
```
dinhviethoa: Please just check the BOOL return value and use NULL for the error pointer (since we don't really use the content).

src/core/zip/MCZip.cpp
```diff
@@ -166,21 +169,32 @@ static ErrorCode addFile(zipFile file, String * path)
...
+    }
     
-    String * tempDirectoryString = String::stringWithFileSystemRepresentation(tempdir);
     String * path = tempDirectoryString->stringByAppendingPathComponent(folder->lastPathComponent())->stringByAppendingUTF8Format(".zip");
     
     ErrorCode err = CreateZipFileFromFolder(path, folder);
     if (err != ErrorNone) {
         unlink(path->fileSystemRepresentation());
-        unlink(tempdir);
+        unlink(tempDirectoryString->UTF8Characters());
```
dinhviethoa: use -fileSystemrepresentation instead of URL8Charcaters()

src/core/zip/MCZip.cpp
```diff
@@ -166,21 +169,32 @@ static ErrorCode addFile(zipFile file, String * path)
...
         return NULL;
     }
```
dinhviethoa: Could you remove the temporaryDirectory here as well (in case of success).

src/core/zip/MCZipMac.mm
dinhviethoa: After a second thought, could you append a [[NSUUID UUID] UUIDString] to the path so we can ensure it's even a unique path for this process? We need to do it because we are removing this folder when the zip file is created.

dinhviethoa:
Build failed: you need to add some changes to src/cmake/core.cmake

check the lines beginning with:

```
IF(APPLE)
```

and how `basetypes_files_apple` is added to the build.

Thanks!

dinhviethoa:
Thanks for shaping the fix! It's now merged.

juliangsp:
Great! you finished the fix :+1:
Thanks for your guidance in the code.

juliangsp:
Do you have a plan for when the next release will come out? Will it be available also via Cocoapods?

PR #1182, opened 2015-07-19
Add nullability annotations

lucasderraugh:
Add nullability annotations to better operate in Swift environments.

### DIFF
--- a/src/objc/imap/MCOIMAPAppendMessageOperation.h
+++ b/src/objc/imap/MCOIMAPAppendMessageOperation.h
@@ -15,6 +15,7 @@
 #import <MailCore/MCOIMAPBaseOperation.h>
 #import <MailCore/MCOConstants.h>
 
+NS_ASSUME_NONNULL_BEGIN
 @interface MCOIMAPAppendMessageOperation : MCOIMAPBaseOperation
 
 @property (nonatomic, assign) time_t date;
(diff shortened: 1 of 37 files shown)


dinhviethoa:
What does NS_ASSUME_NONNULL_BEGIN do?

dinhviethoa:
Also, does it mean that we break with Xcode 6?

lucasderraugh:
https://developer.apple.com/swift/blog/?id=25
NS_ASSUME_NONNULL_BEGIN implies that you cannot pass nil to arguments. The issue with running this in Swift is that the completion handlers are all implicitly unwrapped, which isn't true. If you use any of the values in the completion handler assuming they are are implicitly unwrapped (which is what the current method signature suggests), it will crash your code.

By adding the __nullable modifiers, the compiler suggests that the value is optional, so if you try to print it for example, it will assume it is an optional and will print it as such, instead of crashing.

lucasderraugh:
I'd like to add that to quote from Apple's post 
>What if your Objective-C framework has existing code written against it? Is it safe to just change your types like this? Yes, it is.
So I don't think this can break any backwards compatibility.

dinhviethoa:
Could you investigate and let me know why the build is failing?
https://travis-ci.org/MailCore/mailcore2/jobs/71633120
Do you think it's building using a version of Xcode lower than 6.3?

lucasderraugh:
From Xcode 6.3 release notes, the compiler was bumped to LLVM 6.1, which the tests are running 6.0. I assume this is the problem.

dinhviethoa:
Thanks. I updated the .travis.yml file. Let's see if it passes.

lucasderraugh:
On line 29 of the execution it says `Apple LLVM version 6.0 (clang-600.0.54) (based on LLVM 3.5svn)`. So I assume it is still using LLVM 6.0.

dinhviethoa:
Could you merge with master to set if the build is fixed with my changes?

dinhviethoa:
Great! Thanks!

PR #1197, opened 2015-08-03
Update com_libmailcore_IndexSet.cpp

DaveInAus:
In JNIEXPORT jobject JNICALL Java_com_libmailcore_IndexSet_allRanges
parameter passed is incorrect causing a crash in GetMethodID
    jmethodID method = env->GetMethodID(cls, "add", "(Ljava.lang.Object;)Z");
should be 
    jmethodID method = env->GetMethodID(cls, "add", "(Ljava/lang/Object;)Z");


dinhviethoa:
Thanks for your investigation!
Merging it!

PR #1203, opened 2015-08-07
MCSMTPSession: some MS-Exchange SMTP servers reply with permission pr…

serjepatoff:
Some MS-Exchange SMTP servers reply with permission problem undetectable by mailcore. Catch it and set specific error code instead of common one. 

PS. Prefix "5.7.1" is too common to make decision about permission problems based on it. It can also mean "Unable to relay". That's why I use full error message comparison.

### DIFF
--- a/src/core/smtp/MCSMTPSession.cpp
+++ b/src/core/smtp/MCSMTPSession.cpp
@@ -673,11 +673,11 @@ void SMTPSession::sendMessage(Address * from, Array * recipients, Data * message
     }
     else if (r != MAILSMTP_NO_ERROR) {
         if ((responseCode == 550) && (response != NULL)) {
-            if (response->hasPrefix(MCSTR("5.3.4"))) {
+            if (response->hasPrefix(MCSTR("5.3.4 "))) {
                 * pError = ErrorNeedsConnectToWebmail;
                 goto err;
             }
-            else if (response->hasPrefix(MCSTR("5.7.1 Client does not have permissions to send as this sender"))) {
+            else if (response->hasPrefix(MCSTR("5.7.1 "))) {
                 * pError = ErrorSendMessageNotAllowed;
                 goto err;
             }
--- a/src/core/smtp/MCSMTPSession.cpp
+++ b/src/core/smtp/MCSMTPSession.cpp
@@ -672,17 +672,22 @@ void SMTPSession::sendMessage(Address * from, Array * recipients, Data * message
         goto err;
     }
     else if (r != MAILSMTP_NO_ERROR) {
-        if ((responseCode == 550) && (response != NULL) && (response->hasPrefix(MCSTR("5.3.4")))) {
-            * pError = ErrorNeedsConnectToWebmail;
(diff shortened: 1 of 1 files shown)


src/core/smtp/MCSMTPSession.cpp
```diff
@@ -672,17 +672,22 @@ void SMTPSession::sendMessage(Address * from, Array * recipients, Data * message
...
-            MC_SAFE_REPLACE_COPY(String, mLastSMTPResponse, response);
-            mLastLibetpanError = r;
-            mLastSMTPResponseCode = responseCode;
-            goto err;
+        if ((responseCode == 550) && (response != NULL)) {
+            if (response->hasPrefix(MCSTR("5.3.4"))) {
+                * pError = ErrorNeedsConnectToWebmail;
+                goto err;
+            }
+            else if (response->hasPrefix(MCSTR("5.7.1 Client does not have permissions to send as this sender"))) {
```
dinhviethoa: Could you remove everything that's after "5.7.1 "? Thanks.

src/core/smtp/MCSMTPSession.cpp
```diff
@@ -672,17 +672,22 @@ void SMTPSession::sendMessage(Address * from, Array * recipients, Data * message
...
-            goto err;
-        }
-        else {
-            * pError = ErrorSendMessage;
-            MC_SAFE_REPLACE_COPY(String, mLastSMTPResponse, response);
-            mLastLibetpanError = r;
-            mLastSMTPResponseCode = responseCode;
-            goto err;
+        if ((responseCode == 550) && (response != NULL)) {
+            if (response->hasPrefix(MCSTR("5.3.4"))) {
```
dinhviethoa: Could you take the opportunity to improve that line and add a space after "5.3.4"? Thanks!

serjepatoff: Hello Hoà,

1. Could you remove everything that's after "5.7.1 "?

* "5.7.1 " by itself is too common case to consider it permission-related.
It can also be concerned with SMTP relay issues.
I can check, for example, like this:

if (response->hasPrefix(MCSTR("5.7.1 ")) &&
strstr(response->lowercaseString()->UTF8String(), "permission") !=
NULL)


2. Could you take the opportunity to improve that line and add a space
after "5.3.4"

* For sure

dinhviethoa: According to https://tools.ietf.org/html/rfc3463:

```
3.8 Security or Policy Status

  X.7.1   Delivery not authorized, message refused

         The sender is not authorized to send to the destination.  This
         can be the result of per-host or per-recipient filtering.  This
         memo does not discuss the merits of any such filtering, but
         provides a mechanism to report such.  This is useful only as a
         permanent error.
```

ErrorSendMessageNotAllowed sounds like a good error code to me.
Let me know if you have some concern.

dinhviethoa:
Sorry for the delay.
It looks great! Thanks!

PR #1213, opened 2015-08-24
Include List Converter in javaToMCObject

DaveInAus:
Converter for Java List not included in javaToMCObject. All Java List Parameters in methods will crash as the converter is not found.


dinhviethoa:
Thanks!

PR #1254, opened 2015-10-15
add ability for modular import

avriy:
Add ability for just use import MailCore in swift sources


dinhviethoa:
Could you describe what you did with MailCore.h?
It sounds like you added it to a build, I'm not sure which one.

avriy:
I added it to mailcore osx target and mailcore ios target as a public header

dinhviethoa:
There was already a copy phase script that would copy the headers.
Does adding the headers in the header copy phase help with anything?
What happens if you remove the reference to this header?

avriy:
I know it) Without the reference there is a warning in mailcore osx target: 
warning: no umbrella header found for target 'mailcore osx', module map will not be generated
and then "import MailCore" fails to find module with
No such module 'MailCore'

dinhviethoa:
Sounds good. Thanks!

PR #1266, opened 2015-11-06
Adds support for sending custom IMAP command

### LINKED ISSUE #1247: Custom IMAP command
Hi,

I need to send custom IMAP command in my app (let's say ACTIVATE_CUSTOM_SERVICE) so my communication would go like:

1 STARTTLS
2 CAPABILITY
3 ACTIVATE_CUSTOM_SERVICE

Is there any way to do so? Is there an API to create custom operation which would result in sending this custom command to the server? Thanks.

libec:
Also closes #1247

### DIFF
--- a/src/core/imap/MCIMAPSession.cpp
+++ b/src/core/imap/MCIMAPSession.cpp
@@ -996,9 +996,16 @@ static uint64_t get_mod_sequence_value(mailimap * session)
     return mod_sequence_value;
 }
 
-String * IMAPSession::sendCustomCommand(String * command)
+String * IMAPSession::customCommand(String * command, ErrorCode * pError)
 {
-    mailimap_custom_command(mImap, MCUTF8(command));
+    int r;
+    
+    r = mailimap_custom_command(mImap, MCUTF8(command));
+    if (r == MAILIMAP_ERROR_CUSTOM_COMMAND) {
+        * pError = ErrorCustomCommand;
+        return NULL;
+    }
+    
     String *response = String::stringWithUTF8Characters(mImap->imap_response);
     return response;
 }
--- a/src/core/imap/MCIMAPSession.cpp
+++ b/src/core/imap/MCIMAPSession.cpp
@@ -996,13 +996,11 @@ static uint64_t get_mod_sequence_value(mailimap * session)
     return mod_sequence_value;
 }
 
-void IMAPSession::sendCustomCommand(String * command)
+String * IMAPSession::sendCustomCommand(String * command)
 {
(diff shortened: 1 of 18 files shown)


libec:
I'd say it fails because mailcore first tries to download prebuilt libetpan, right?

dinhviethoa:
It think that's the issue. I'll update it this week-end.
Meanwhile, could you update the CMake build file? It should be easy enough even if you don't know yet about CMake. It's mainly in src/cmake.

libec:
Sure, will do tomorrow. Have to go afk right now.

src/core/imap/MCIMAPSession.cpp
```diff
@@ -996,6 +996,13 @@ static uint64_t get_mod_sequence_value(mailimap * session)
     return mod_sequence_value;
 }
 
+String * IMAPSession::sendCustomCommand(String * command)
```
dinhviethoa: s/sendCustomCommand/customCommand/
Also, you need to add an ErrorCode parameter.

libec:
I've added the `ErrorCode` as parameter to `customCommand` calls, please let me know if you think it's handled properly or needs some more work. I've also changed the name of the functions from `sendCustomCommand` to `customCommand`.

src/core/imap/MCIMAPSession.cpp
```diff
@@ -996,6 +996,20 @@ static uint64_t get_mod_sequence_value(mailimap * session)
     return mod_sequence_value;
 }
 
+String * IMAPSession::customCommand(String * command, ErrorCode * pError)
+{
+    int r;
+    
+    r = mailimap_custom_command(mImap, MCUTF8(command));
+    if (r == MAILIMAP_ERROR_CUSTOM_COMMAND) {
```
dinhviethoa: Could you change the error management to the following?

```
    if (r == MAILIMAP_ERROR_STREAM) {
        mShouldDisconnect = true;
        * pError = ErrorConnection;
        return;
    }
    else if (r == MAILIMAP_ERROR_PARSE) {
        * pError = ErrorParse;
        return;
    }
    else if (hasError(r)) {
        * pError = ErrorDelete;
        return;
    }
```

dinhviethoa:
One comment then we're good to go.

dinhviethoa:
I'll fixed it after merge.

libec:
Thanks for your guidance :+1: Great work on libetpan and mailcore btw. !

dinhviethoa:
Thanks for your patch!

PR #1331, opened 2016-01-10
iOS build improvements

conradev:
I made two changes:

1. I filled out the iOS framework deployment version to be iOS 8.0, which is the minimum OS version that dynamic frameworks can be deployed to on iOS
2. I enabled the `Allow app extension API only` option because the framework only uses APIs that are safe for use in extensions


dinhviethoa:
Thanks!

PR #1340, opened 2016-01-19
Make a NULL OAUTH Token an STMP Stream Error

### LINKED ISSUE #1330: MCOAuthTypeXOAuth2 Causes Crash iOS9
Following the example outlined here:
https://github.com/MailCore/mailcore2/wiki/SMTP-Examples

I found that my app was considered not up to snuff by today's security standards according to google.

My first response was to begin declaring the authorization type for the `smtpSession` object.

I cannot seem to use `MCOAuthTypeXOAuth2` because it causes my application to crash while sending the email.

All other types seem to work just fine, but don't grant me secure access to GMail's SMTP.

I'm forced to allow the GMail address to give access to "Low Security Applications", which sounds awful.

This is on iOS9.

CodaFi:
Fixes #1330 by making the code mirror its `IMAP` counterpart.


dinhviethoa:
Thanks!

dinhviethoa:
And welcome back :)

CodaFi:
@dinhviethoa Glad to be back!

Issue #1310, opened 2015-12-12
Mailcore dealloc fails


KennyVB:
![skaermbillede 2015-12-12 kl 13 24 26](https://cloud.githubusercontent.com/assets/727549/11761912/ff954542-a0d3-11e5-9ad2-34e91c46475e.png)

KennyVB:
this is what happens when i close a web view AFTER viewing a mail

dinhviethoa:
Please, create a small project on github that reproduces the issue.
Thanks.

KennyVB:
https://github.com/KennyVB/mailcore2test

dinhviethoa:
If you initialize that to nil / null instead of MCOIMAPMessageRenderingOperation(), it will probably solve the crash. I'm not sure what you intent was here.
https://github.com/KennyVB/mailcore2test/blob/master/Blackmail/tableViewController.swift#L21

KennyVB:
hmm now i tried removing it altogether... still crashes

dinhviethoa:
Could you show the new crash log?

CodaFi:
It's because `MCOOperation` has no designated `init` (probably should start annotating these APIs with that soon), yet your rendering view controller is calling it.  When your view gets popped, you'll destroy the instance you created and cause the operation to be over-released.  I don't know if this is a user error or if we should check for null in the destructor.

dinhviethoa:
Could you create a pull request with those annotations?
Thanks!

CodaFi:
I can.  But more importantly should the init() be marked unavailable?

dinhviethoa:
what do you mean?

CodaFi:
Well, MCOOperation doesn't look like it's setup for that call to be legal unless it's being delegated from a subclass.  So we can mark it deprecated or unavailable to prevent these kinds of things from happening in the first place, or implement -init and allow it but fix the crash.

dinhviethoa:
Maybe a -init with a big assert would just work.

PR #1341, opened 2016-01-19
Add Designated Initializers

CodaFi:
Adds designated initializer marks to explicit init methods per the discussion in #1310.  Also makes some initializers unavailable to prevent the kinds of errors in there too.

### DIFF
--- a/src/objc/abstract/MCOAbstractMessage.h
+++ b/src/objc/abstract/MCOAbstractMessage.h
@@ -56,4 +56,8 @@ namespace mailcore {
 
 @end
 
+@interface MCOAbstractMessage (MCOUnavailable)
+- (id) init NS_UNAVAILABLE;
+@end
+
 #endif
--- a/src/objc/abstract/MCOAbstractMessage.h
+++ b/src/objc/abstract/MCOAbstractMessage.h
@@ -57,7 +57,10 @@ namespace mailcore {
 @end
 
 @interface MCOAbstractMessage (MCOUnavailable)
-- (id) init NS_UNAVAILABLE;
+
+/** Do not invoke this directly */
+- (id) init;
+
 @end
 
 #endif
--- a/src/objc/abstract/MCOAbstractMessage.mm
+++ b/src/objc/abstract/MCOAbstractMessage.mm
@@ -18,7 +18,7 @@
 #import "NSString+MCO.h"
 
 @interface MCOAbstractMessage ()
-- (id)init NS_DESIGNATED_INITIALIZER;
+- (id) init NS_DESIGNATED_INITIALIZER;
 @end
 
 @implementation MCOAbstractMessage {
--- a/src/objc/abstract/MCOAbstractMessage.mm
+++ b/src/objc/abstract/MCOAbstractMessage.mm
@@ -17,10 +17,6 @@
 #import "NSObject+MCO.h"
 #import "NSString+MCO.h"
 
-@interface MCOAbstractMessage ()
-- (id) init NS_DESIGNATED_INITIALIZER;
-@end
-
 @implementation MCOAbstractMessage {
     mailcore::AbstractMessage * _message;
 }
@@ -34,9 +30,9 @@
 
 - (id) init
 {
+    self = [self initWithMCMessage:NULL];
     MCAssert(0);
-    [self release];
-    return (self = nil);
+    return nil;
 }
 
 - (id) initWithMCMessage:(mailcore::AbstractMessage *)message
--- a/src/objc/abstract/MCOAbstractMessage.mm
+++ b/src/objc/abstract/MCOAbstractMessage.mm
@@ -17,6 +17,10 @@
 #import "NSObject+MCO.h"
 #import "NSString+MCO.h"
 
+@interface MCOAbstractMessage ()
+- (id)init NS_DESIGNATED_INITIALIZER;
+@end
+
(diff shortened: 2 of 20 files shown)


CodaFi:
What to do about MCOMessageBuilder?  I want to keep `-init` unavailable, but it seems required.  Maybe a new convenience initializer?

CodaFi:
There, compromise.  I think I'm going to continue modernizing in other pull requests, unless there's anybody that still needs the Objective-C part of this framework to compile with GCC or something.

dinhviethoa:
MCOMessageBuilder needs init. When we want to build messages from scratch, that's useful!

CodaFi:
Yep, I just put it back and made the abstract class throw instead.

src/objc/abstract/MCOAbstractMessage.h
```diff
@@ -56,4 +56,11 @@ namespace mailcore {
 
 @end
 
+@interface MCOAbstractMessage (MCOUnavailable)
+
+/** Do not invoke this directly */
+- (id) init;
```
dinhviethoa: Why don't you use NS_UNAVAILABLE here?

src/objc/abstract/MCOAbstractPart.h
```diff
@@ -109,4 +109,8 @@ namespace mailcore {
 
 @end
 
+@interface MCOAbstractPart (MCOUnavailable)
+- (id) init NS_UNAVAILABLE;
```
CodaFi: I couldn't find any `MCOAbstractPart` subclasses that needed `-init` called.  If you know of any, please let me know and I'll remove this annotation.

src/objc/abstract/MCOAbstractMessage.h
CodaFi: Because I can't make MCOMessageBuilder's `-init` available again in a subclass.

src/objc/rfc822/MCOMessageParser.mm
```diff
@@ -16,6 +16,10 @@
 #import "MCOUtils.h"
 #import "MCOAbstractMessageRendererCallback.h"
 
+@interface MCOMessageParser ()
```
dinhviethoa: This is weird...

src/objc/abstract/MCOAbstractMessage.mm
```diff
@@ -17,6 +17,10 @@
 #import "NSObject+MCO.h"
 #import "NSString+MCO.h"
 
+@interface MCOAbstractMessage ()
+- (id)init NS_DESIGNATED_INITIALIZER;
```
dinhviethoa: This is weird... (inside the implementation)

src/objc/abstract/MCOAbstractMessage.mm
```diff
@@ -30,9 +34,9 @@ @implementation MCOAbstractMessage {
 
 - (id) init
 {
-    self = [self initWithMCMessage:NULL];
     MCAssert(0);
-    return nil;
+    [self release];
```
dinhviethoa: You can keep the previous code. MCAssert() is making sure that it's not used and crashing.

src/objc/abstract/MCOAbstractMessage.mm
CodaFi: Yeah... It has to be a part of the main interface or part of an anonymous class extension (here) for it to count.  I didn't want to put it in the main interface because it sends the wrong message, but I can move it back if you want.

dinhviethoa: Therefore, don't make that change at all if the language doesn't really help.

dinhviethoa:
Could you start with couple of `-init NS_UNAVAILABLE` only?

src/objc/abstract/MCOAbstractMessage.mm
CodaFi: Can do :+1:

src/objc/rfc822/MCOMessageParser.mm
CodaFi: Necessary because the initializer is declared in a private extension and cannot be marked designated

CodaFi:
I can remove them from this pull request and factor them into another.  The bulk of the work in here is the designated init parts.

dinhviethoa:
Could you rebase that pull request and squash the commits?

dinhviethoa:
@CodaFi Should I close this pull request?

CodaFi:
Yeah, I think it's for the best.

CodaFi:
Do you know if there's any outstanding issues with compatibility on the Objective-C end of this?  Before I start improving Swift stuff, I want to try modernizing everything (nullability, instancetype, generics, etc.) and if we have people using GCC or something then that's going to cause a lot of problems.

dinhviethoa:
The things I heard of were the delegates names, for example, for the message renderer.

CodaFi:
What I mean is, do you know how far back we need to target in terms of compilers for the Objective-C stuff: are we requiring the latest LLVM toolchain for it?

dinhviethoa:
The codebase compiles on Visual Studio.
For Obj-C, can we make transition smooth?

CodaFi:
If we're targeting VS2015, then it shouldn't be a problem.  I'll just make the changes and we'll see what happens, eh?

PR #1343, opened 2016-01-27
2 new error codes

serjepatoff:
Hello,

SMTP server smtp.mail.yahoo.co.jp can reply with 2 custom errors while trying to send mail:

*554 Cannot send message due to possible spam; please visit https://www.yahoo-help.jp/app/ask/p/2494/form/betamail-inquiry for more information*

and

*521 smtp.mail.yahoo.co.jp closing transmission channel. User is over the limit for messages allowed to be sent in a single day*.

MCSMTPSession doesn't treat them and returns common error code 'ErrorSendMessage' that does not reflect the exact reason of fault.

To be able to handle these error cases in our email client (based on mailcore2), I have added two custom error codes.

Can you add them to the mainline mailcore2?

Thank you very much for your attention!

PS. Also I've found and fixed potential bug in MCErrorMessage::errorMessageWithErrorCode() that returns NULL when fed with recently added ErrorCustomCommand code.

### DIFF
--- a/src/core/smtp/MCSMTPSession.cpp
+++ b/src/core/smtp/MCSMTPSession.cpp
@@ -682,12 +682,12 @@ void SMTPSession::sendMessage(Address * from, Array * recipients, Data * message
                 goto err;
             }
         }
-        else if (responseCode == 521 && response->locationOfString(MCSTR("limit")) != -1) {
-            * pError = ErrorSendMessageDailyLimitExceeded;
+        else if (responseCode == 521 && response->locationOfString(MCSTR("over the limit")) != -1) {
+            * pError = ErrorYahooSendMessageDailyLimitExceeded;
             goto err;
         }
         else if (responseCode == 554 && response->locationOfString(MCSTR("spam")) != -1) {
-            * pError = ErrorSendMessageSpamSuspected;
+            * pError = ErrorYahooSendMessageSpamSuspected;
             goto err;
         }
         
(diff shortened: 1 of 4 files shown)


src/core/smtp/MCSMTPSession.cpp
```diff
@@ -682,6 +682,14 @@ void SMTPSession::sendMessage(Address * from, Array * recipients, Data * message
                 goto err;
             }
         }
+        else if (responseCode == 521 && response->locationOfString(MCSTR("limit")) != -1) {
```
dinhviethoa: Can you check for "over the limit" instead?

src/objc/abstract/MCOConstants.h
```diff
@@ -413,6 +413,10 @@ typedef NS_ENUM(NSInteger, MCOErrorCode) {
     MCOErrorNoValidServerFound,
     /** Error while running custom command */
     MCOErrorCustomCommand,
+    /** Spam was suspected by server */
+    MCOErrorSendMessageSpamSuspected,
```
dinhviethoa: Could you prefix those error messages with Yahoo?

See:
MCOErrorGmailTooManySimultaneousConnections
or
MCOErrorYahooUnavailable

dinhviethoa:
That's a great addition! Thanks for sending this.
It's almost ready to merge. I added some comments.

dinhviethoa:
Thanks a lot!

PR #1346, opened 2016-02-02
Reduce memory usage when fetch message body by IMAP

disaykin:


disaykin:
I test this changes on my data. But I could not find the official test set to validate the changes.

src/core/basetypes/MCData.cpp
```diff
@@ -510,44 +534,47 @@ Data * Data::dataWithContentsOfFile(String * filename)
...
     }
-    
-    data = Data::data();
-    data->takeBytesOwnership(buf, (unsigned int) stat_buf.st_size);
-    
-    fclose(f);
-    
-    return data;
+
+    Data * result = new Data();
```
dinhviethoa: Could you go back to:
```
data = new Data::dataWithMmapedData(bytes, length);
```

```
Data * dataWithMmapedData()
{
  data = Data::data();
  data->takeBytesOwnership(buf, (unsigned int) stat_buf.st_size, mmapDeallocator);
}
```

Could you also add a readonly flag to the data when takeBytesOwnership() is used?

And then, we need some fallback implementation, which would make Data back to
malloc'ed blocks when the user of the API tries to call one of the following methods:

```
        virtual void appendData(Data * otherData);
        virtual void appendBytes(const char * bytes, unsigned int length);
        virtual void setBytes(const char * bytes, unsigned int length);
        virtual void setData(Data * otherData);
```

src/core/basetypes/MCData.cpp
```diff
@@ -606,30 +636,27 @@ Data * Data::decodedDataUsingEncoding(Encoding encoding)
...
             data = Data::dataWithCapacity((unsigned int) text_length);
-            
-            dup_data = (char *) malloc(text_length);
-            memcpy(dup_data, text, text_length);
-            
-            current_p = dup_data;
+
+            current_p = text;
             while (1) {
+                char decoded_buf[256];
```
dinhviethoa: I think there might be some issue when the buffer is too small.

src/core/basetypes/MCData.cpp
```diff
@@ -664,23 +691,22 @@ Data * Data::decodedDataUsingEncoding(Encoding encoding)
                 if (length == 0) {
                     break;
                 }
-                decoded_length = uudecode(current_p, length);
+                decoded_length = uudecode(current_p, length, decoded_buf, sizeof(decoded_buf));
```
dinhviethoa: Did you notice big memory usage is real use case for uudecode()?
It doesn't seem like an encoding format that's used a lot.
I would be fine with keeping the previous implementation of the decoding uuencoded parts.

src/core/basetypes/MCData.h
```diff
@@ -28,7 +30,13 @@ namespace mailcore {
         static Data * dataWithCapacity(int capacity);
         static Data * dataWithContentsOfFile(String * filename);
         static Data * dataWithBytes(const char * bytes, unsigned int length);
-        
+
+         /* Create instance of Data with contents of given bytes without extra copying.
+         Memory ownership transferred from client code to the Data instance.
+         Data destructor will call bytesDeallocator function for memory releasing.
+         */
+        static Data * dataWithBytesTakeOwnership(char * bytes, unsigned int length, BytesDeallocator bytesDeallocator);
```
dinhviethoa: I would make it private.

src/core/basetypes/MCData.cpp
disaykin: Yes, I can. I do it tomorrow.
Fallback works via calls of reset() and allocate() now. I will rewrite patch to make it more clear.

src/core/basetypes/MCData.cpp
disaykin: In UUE-encoding all data lines except the last (if the data was not divisible by 45), have 45 bytes of encoded data (60 characters after encoding). So I think 255-bytes buffer is enough.

src/core/basetypes/MCData.cpp
disaykin: I work on iOS imap client. iOS app has very strong memory constraints. The app may be killed if it uses a lot of memory. UUE is very unpopular encoding format, but I think it is very bad explanation for user why application was crashed.

src/core/basetypes/MCData.h
disaykin: This method is used in some other places. For example, in 
Data * IMAPSession::fetchMessageAttachment() method.

dinhviethoa: ok, could you move it to public: // private with charsetWithFilteredHTML?

src/core/basetypes/MCData.cpp
dinhviethoa: Ok. since uu encoding format is not too popular, I guess we can fail if the encoding is wrong (more than 200 bytes / lines, etc). Could you add some comment to explain that uu encoding is usually 45 bytes or less / line?

src/core/basetypes/MCData.cpp
```diff
@@ -42,6 +43,18 @@ void Data::allocate(unsigned int length, bool force)
     if (length <= mAllocated)
         return;
 
+    if (mBytesDeallocator) {
+        if (mBytes) {
+            unsigned int bytes_len = mLength;
+            char * bytes = strndup(mBytes, mLength);
```
dinhviethoa: Please don't use strndup() since it won't behave correctly if there's a \0 in the data.
use malloc() and memcpy().

dinhviethoa:
That's an awesome improvement. I can't wait to get this patch in!

disaykin:
@dinhviethoa I hope I have corrected all your comments.

dinhviethoa:
Thanks a lot for this improvement!

disaykin:
Thanks!

PR #1351, opened 2016-02-05
Imap move

kolyuchiy:
This change also requires the most recent changes from libetpan. I'm not sure how to properly update this dependency in this repository.


dinhviethoa:
Awesome! :8ball: 
Let me build an updated libetpan, check the build. Then, I'll merge it.

Issue #1352, opened 2016-02-08
CocoaPods: Undefined symbols for architecture


skofgar:
Hi, I just added **mailcore2** to a new project and cannot get it running. I always get this error:
```
Undefined symbols for architecture x86_64:
  "_res_9_dn_expand", referenced from:
      mailcore::MXRecordResolverOperation::main() in libMailCore.a(MCMXRecordResolverOperation.o)
  "_res_9_ns_initparse", referenced from:
      mailcore::MXRecordResolverOperation::main() in libMailCore.a(MCMXRecordResolverOperation.o)
  "_res_9_ns_parserr", referenced from:
      mailcore::MXRecordResolverOperation::main() in libMailCore.a(MCMXRecordResolverOperation.o)
  "_res_9_ns_sprintrr", referenced from:
      mailcore::MXRecordResolverOperation::main() in libMailCore.a(MCMXRecordResolverOperation.o)
  "_res_9_search", referenced from:
      mailcore::MXRecordResolverOperation::main() in libMailCore.a(MCMXRecordResolverOperation.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

My Podfile looks like this:

```
platform :osx, '10.11'
use_frameworks!

target 'MyProject' do
  pod 'mailcore2-osx'
end
```

skofgar:
This could be the answer:

I simply added following in the _main target_ of my project → `Build Settings` →  `Other Linker Flags`: 
```
-lresolv
```

suculent:
Wow, thanks.

dinhviethoa:
Could you try the new cocoapods? It shouldn't require you to go through this step.

skofgar:
@dinhviethoa Yes it worked for me. Thank you!

PR #1353, opened 2016-02-08
Added suggested linker flags to cocoapods-section

skofgar:
Added further steps to Cocoapods-section to solve `Undefined symbols for arm64` build-error. To prevent this from happening `-lresolv` needs to be added to the `Other Linker Flags`.
Related issue: #1352

### DIFF
--- a/build-mac/README.md
+++ b/build-mac/README.md
@@ -25,6 +25,11 @@ pod 'mailcore2-ios'
 pod 'mailcore2-osx'
 ```
 
+**Further Steps**
+
+You may also need to add `-lresolv` to your application-target -> `Build Settings` -> `Other Linker Flags`.
+
+
 ### Binary ###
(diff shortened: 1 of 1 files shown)


dinhviethoa:
Actually, I think we need to solve the cocoapods podspec.
Could you check whether adding `resolv` to that list in the podspec would solve the issue?
```
  "libraries": ["xml2", "iconv", "z", "c++"],
```
(see mailcore2-ios.podspec.json)

dinhviethoa:
ref #1352.

dinhviethoa:
I fixed the cocoapods instead. Let me know if it doesn't work for you.

skofgar:
@dinhviethoa The update seems to work for me now – thanks! (I didn't get to try updating the podspecs...)

PR #1359, opened 2016-02-09
Mark non-designated initializers and new as unavailable

kolyuchiy:


dinhviethoa:
There's a build failure on Mac/iOS. Could you take a look?
Thanks!

PR #1363, opened 2016-02-15
Added ability to customize the identity of IMAP client in the time of…

disaykin:
… session initialization.

The Mailcore2 will use it in loginIfNeeded() instead of sending "1 ID NIL".


dinhviethoa:
The previous way to do it was to modify the clientIdentity.
Anyway, people wanted to use that API this way.
So that's great!

dinhviethoa:
Thanks!

PR #1367, opened 2016-02-23
Add upload functionality to NNTP sessions.

CTMacUser:
> Add new methods to core, async, and objc NNTP sessions to upload
> message data. For the async and objc sessions, add classes to wrap
> upload operations.

I created the code by copying the corresponding operation in the SMTP code, then removing the sender and recipient parameters (needed in SMTP since the parameters are submitted in piecemeal during the protocol to support Bcc), and then copying one of the Fetch-Article operation's code to fill the holes. There may be other particulars that I have missed.

There are no new tests for this code. I just used Xcode, so there are no accommodations for other platforms (It probably works for iOS, though.), including any makefiles. I copied-and-pasted the code to strip the Bcc section of a message; maybe that can be factored out to a new code file. (I don't know where though.)

### DIFF
--- a/src/objc/nntp/MCONNTPSendOperation.h
+++ b/src/objc/nntp/MCONNTPSendOperation.h
@@ -1,0 +1,41 @@
+//
+//  MCONNTPSendOperation.h
+//  mailcore2
+//
+//  Created by Daryle Walker on 2/21/16.
+//  Copyright © 2016 MailCore. All rights reserved.
+//
+
+#ifndef MAILCORE_MCONNTPSENDOPERATION_H
+
+#define MAILCORE_MCONNTPSENDOPERATION_H
+
+#import <Foundation/Foundation.h>
+#import <MailCore/MCONNTPOperation.h>
+
+/** Send a message from NNTP3 */
+
+typedef void (^MCONNTPOperationProgressBlock)(unsigned int current, unsigned int maximum);
+
+NS_ASSUME_NONNULL_BEGIN
+@interface MCONNTPSendOperation : MCONNTPOperation
+
+/** This block will be called as data is downloaded from the network */
+@property (nonatomic, copy) MCONNTPOperationProgressBlock progress;
+
+/**
+ Starts the asynchronous send operation.
+ 
+ @param completionBlock Called when the operation is finished.
+ 
+ - On success `error` will be nil
+ 
+ - On failure, `error` will be set with `MCOErrorDomain` as domain and an
+ error code available in MCOConstants.h
+ */
+- (void) start:(void (^)(NSError * __nullable error))completionBlock;
+
+@end
+NS_ASSUME_NONNULL_END
+
+#endif
--- a/src/objc/nntp/MCONNTPSendOperation.h
+++ b/src/objc/nntp/MCONNTPSendOperation.h
@@ -13,10 +13,6 @@
 #import <Foundation/Foundation.h>
 #import <MailCore/MCONNTPOperation.h>
 
-/** Send a message from NNTP3 */
-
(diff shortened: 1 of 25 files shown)


CTMacUser:
I created the code by copying the corresponding operation in the SMTP code, then removing the sender and recipient parameters (needed in SMTP since the parameters are submitted in piecemeal during the protocol to support Bcc), and then copying one of the Fetch-Article operation's code to fill the holes. There may be other particulars that I have missed.

There are no new tests for this code. There are no accommodations for other platforms besides OS X (It probably works for iOS, though.), including any makefiles. I copied-and-pasted the code to strip the Bcc section of a message; maybe that can be factored out to a new code file. (I don't know where though.)

dinhviethoa:
Could you fix the cmake build? Thanks!

src/objc/nntp/MCONNTPSendOperation.h
```diff
@@ -0,0 +1,41 @@
...
+#ifndef MAILCORE_MCONNTPSENDOPERATION_H
+
+#define MAILCORE_MCONNTPSENDOPERATION_H
+
+#import <Foundation/Foundation.h>
+#import <MailCore/MCONNTPOperation.h>
+
+/** Send a message from NNTP3 */
+
+typedef void (^MCONNTPOperationProgressBlock)(unsigned int current, unsigned int maximum);
```
dinhviethoa: Please move it to MCONNTPOperation since the same is defined in NNTPFetchArticleOperation.

src/objc/nntp/MCONNTPSession.mm
```diff
@@ -205,6 +206,17 @@ - (MCONNTPListNewsgroupsOperation *) listDefaultNewsgroupsOperation {
     return MCO_TO_OBJC_OP(coreOp);
 }
 
+- (MCONNTPSendOperation *) sendOperationWithData:(NSData *)messageData {
```
dinhviethoa: Maybe postOperationWithData: is more appropriate?

src/objc/nntp/MCONNTPSession.mm
```diff
@@ -205,6 +206,17 @@ - (MCONNTPListNewsgroupsOperation *) listDefaultNewsgroupsOperation {
...
+    mailcore::NNTPSendOperation * coreOp = MCO_NATIVE_INSTANCE->sendMessageOperation(MCO_FROM_OBJC(mailcore::Data, messageData));
+    return MCO_TO_OBJC_OP(coreOp);
+}
+
+- (MCONNTPSendOperation *) sendOperationWithContentsOfFile:(NSString *)path
```
dinhviethoa: same post here.

dinhviethoa:
Thanks a lot for this code!

dinhviethoa:
Thanks! That's a great addition to NNTP.

dinhviethoa:
Are you working on a cool news client? :)

CTMacUser:
I wonder if `MCONNTPOperationProgressBlock` should be moved again, this time to one of those headers that contain a bunch of type definitions. And the fetch-article method never uses the callback (because there was never one to use for connections until I added it).

dinhviethoa:
I'm not sure ... why is it ok in src/objc/nntp/MCONNTPOperation.h?

CTMacUser:
I thought that was your original suggestion. I later saw that "MCOConstants.h" not only has various `enum` definitions, but a couple of function-pointer types. I realize that maybe this header is a better place, to reduce the potential number of headers a user has to search. (Well, Xcode can automatically jump to a type's definition, so I don't think this concern is major.)

dinhviethoa:
That's how IMAP is right now. If you want to do that, they have to be consistent.
But don't worry about it too much.

PR #1368, opened 2016-02-23
Fixed html renderer

foxinushka:


dinhviethoa:
Could you describe the issue you had?

src/core/renderer/MCHTMLRenderer.cpp
```diff
@@ -464,8 +464,9 @@ static String * htmlForAbstractMultipartMixed(AbstractMultipart * part, htmlRend
         AbstractPart * subpart = (AbstractPart *) part->parts()->objectAtIndex(i);
         String * substring = htmlForAbstractPart(subpart, context);
         if (context->pass != 0) {
-            if (substring == NULL)
-                return NULL;
+            if (substring == NULL) {
```
CodaFi: If you're getting null here, it indicates a failure in `htmlForAbstractPart`.  I don't think this should be the workaround for that kind of logic error.

foxinushka:
I found where I've mistaken.

I have a message with next structure:

```
multipart
   multipart
      text/plain
      text/html
   text/plain (inline attachment)
```

When I render HTML content of the message, I should skip text/plain (inline attachment) part.
I skip it in shouldShowPart method of MCOHTMLRendererDelegate. But the method isn't be called for text/... parts, and my code contained call of shouldShowPart in dataForIMAPPart. So, It is my mistake. 

Please, check another fix. 
Thanks!

dinhviethoa:
I'm closing that issue then.

PR #1371, opened 2016-02-29
Close IMAP connection in case of parse error

disaykin:


dinhviethoa:
Could you take a look at how it's done for socket errors?
https://github.com/MailCore/mailcore2/blob/master/src/core/imap/MCIMAPSession.cpp#L1652
We could rather do something similar for parse errors.

disaykin:
Done

src/async/imap/MCIMAPOperation.cpp
```diff
@@ -110,14 +110,18 @@ void IMAPOperation::start()
         IMAPAsyncConnection * connection = mMainSession->sessionForFolder(mFolder, mUrgent);
         setSession(connection);
     }
-    mSession->runOperation(this);
+    session()->runOperation(this);
 }
 
+namespace {
```
dinhviethoa: I know that is a good best practice in C++, though could you remove anyway that anonymous namespace anyway?

dinhviethoa:
One small change. Everything else looks good!

disaykin:
Done.

dinhviethoa:
Thanks!

PR #1379, opened 2016-03-14
Move inttypes include within ifdef __cplusplus in MCIndexSet.h

conradev:
This fixes a "include of non-modular header inside framework" error that appears in Xcode 7.3 and later

### DIFF
--- a/src/core/basetypes/MCIndexSet.h
+++ b/src/core/basetypes/MCIndexSet.h
@@ -12,9 +12,10 @@
 
 #include <MailCore/MCObject.h>
 #include <MailCore/MCRange.h>
-#include <inttypes.h>
 
 #ifdef __cplusplus
+
+#include <inttypes.h>
 
 namespace mailcore {
     


dinhviethoa:
Can you use the setting "Allow Non-modular includes in Framework Modules" instead?
It doesn't seem like the correct way to fix the issue.

conradev:
The way I interpreted it, at least, was that including MailCore as a module from an Objective-C file does not see `inttypes.h` as a valid module, whereas Objective-C++ recognizes it as a valid module.

`MCRange.h` includes `inttypes.h` in this manner and does not raise any errors.

conradev:
Now that Xcode 7.3 is public, would it be possible to merge this?

Another way of phrasing this PR is "Making `MCIndexSet.h` and `MCRange.h` consistent on the include of `inttypes.h`".

dinhviethoa:
I'm unable to reproduce the issue. Could you tell me how you get the build error?

conradev:
Download [this](https://github.com/MailCore/mailcore2/files/185236/MailCoreTest.zip) project, run the following:

```
$ cd MailCoreTest
$ brew install carthage
$ carthage update
```

and then try to build it in Xcode 7.3. You should get an error. If you apply the patch from the pull request, then it will build successfully.

dinhviethoa:
It's probably the best way to workaround the issue :(

PR #1381, opened 2016-03-16
Use MCOIMAPCheckAccountOperation as return type for ObjC IMAP Session::checkAccount operation

disaykin:


dinhviethoa:
I'm not sure why you need to return an explicit type?

disaykin:
I need information about loginResponse. This information returns only via userInfo field now. And only in AccountCheckOperation callback.

Without my changes i don't receive this field in error's userInfo.

But now I think I'm tested in the previous build of the library (without your latest additions of AccountCheckOperation functionality).

So I revert one commit.

src/objc/imap/MCOIMAPCheckAccountOperation.mm
```diff
@@ -62,8 +62,9 @@ - (void) operationCompleted
     } else {
         NSError * error = [NSError mco_errorWithErrorCode:op->error()];
         if (op->loginResponse() != NULL) {
-            NSDictionary * userInfo = @{@"IMAPServerError": MCO_TO_OBJC(op->loginResponse())};
-            error = [NSError errorWithDomain:[error domain] code:[error code] userInfo:userInfo];
+            NSMutableDictionary * userInfo = [[error userInfo] mutableCopy];
+            userInfo[@"IMAPServerError"] = MCO_TO_OBJC(op->loginResponse());
+            error = [NSError errorWithDomain:[error domain] code:[error code] userInfo:[userInfo copy]];
```
dinhviethoa: use `error = [NSError errorWithDomain:[error domain] code:[error code] userInfo:userInfo];` instead.

dinhviethoa:
one minor change and then, it's ready to merge.

disaykin:
Done

Issue #1145, opened 2015-06-18
mIdleEnabled and mCanIdle in MCIMAPSession.cpp


yuklai:
Both seem to mean the same thing. Why are there two of them?
Also why is semaphore used while setting `mCanIdle` but not `mIdleEnabled` (or any of the other capabilities for that matter)?

See `capabilitySetWithSessionState` and `applyCapabilities`

CodaFi:
Fixed by the merge of #1384

Thanks for reporting this!

PR #1384, opened 2016-03-21
Remove duplicate `mCanIdle`

CodaFi:
#1145 raised an interesting point.  I think at one point `mCanIdle` was supposed to stand for a client-side value for the IDLE capability while `mIdleEnabled` was supposed to represent the server-side capability, but somewhere along the way `mCanIdle`'s role got reduced to what it does now which is essentially shadow `mIdleEnabled`.  Far as I can tell this won't have an impact on the semantics of the code and may actually clarify the implementation of `isIdleEnabled()`.

### DIFF
--- a/src/core/imap/MCIMAPSession.cpp
+++ b/src/core/imap/MCIMAPSession.cpp
@@ -344,7 +344,6 @@ void IMAPSession::init()
     mLastFetchedSequenceNumber = 0;
     mCurrentFolder = NULL;
     pthread_mutex_init(&mIdleLock, NULL);
-    mCanIdle = true;
     mState = STATE_DISCONNECTED;
     mImap = NULL;
     mProgressCallback = NULL;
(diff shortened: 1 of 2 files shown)


src/core/imap/MCIMAPSession.cpp
```diff
@@ -3853,7 +3852,7 @@ void IMAPSession::capabilitySetWithSessionState(IndexSet * capabilities)
     }
     if (mailimap_has_idle(mImap)) {
```
dinhviethoa: Can you removed those lines and do that instead in applyCapabilities()?

dinhviethoa: Can you also lock mIdleEnabled line 3919? We need to make sure it's safe.

CodaFi: If the write is locked, shouldn't the accessor be too?

dinhviethoa: You're totally right.

PR #1392, opened 2016-03-29
Added ability of canceling for active sending operation

AlexKar:


src/core/smtp/MCSMTPSession.cpp
```diff
@@ -636,6 +656,13 @@ void SMTPSession::sendMessage(Address * from, Array * recipients, Data * message
         goto err;
     }
 
+    CANCEL_LOCK();
+    sendingCancelled = mSendingCancelled;
+    CANCEL_UNLOCK();
+    if (mSendingCancelled) {
```
dinhviethoa: Please check sendingCancelled instead of mSendingCancelled.

src/core/smtp/MCSMTPSession.cpp
```diff
@@ -864,6 +894,19 @@ void SMTPSession::noop(ErrorCode * pError)
     }
 }
 
+void SMTPSession::cancelMessageSending()
+{
+    // main thread
```
dinhviethoa: mSmtp needs to be locked.

1/ We should perform a mailstream_cancel() only during mailesmtp_send().
2/ One issue is that mailesmtp_send_quit() will change mSmtp->stream.
Therefore, we need to make some changes in mailsmtp to add mailesmtp_send_quit_no_disconnect().

Here's the pseudo-code:

```
sendMessage() {
  lock()
  mCanCancel = true;
  unlock()
  mailesmtp_send()
  lock()
  mCanCancel = false;
  unlock()
}

cancelSend() {
  lock()
  if (mCanCancel) {
    mailstream_cancel()
  }
  unlock()
}
```

AlexKar: done

src/core/smtp/MCSMTPSession.cpp
AlexKar: done

src/async/smtp/MCSMTPSendWithDataOperation.cpp
```diff
@@ -84,3 +84,9 @@ void SMTPSendWithDataOperation::main()
     }
     setError(error);
 }
+
+void SMTPSendWithDataOperation::cancel()
+{
+    SMTPOperation::cancel();
+    session()->session()->cancelMessageSending();
```
dinhviethoa: Could you get rid of setSendingCancelled()?
and check in main() if the operation is cancelled before running sendMessage()?

src/core/smtp/MCSMTPSession.cpp
```diff
@@ -623,6 +649,10 @@ void SMTPSession::sendMessage(Address * from, Array * recipients, Data * message
         return;
     }
     
+    CAN_CANCEL_LOCK();
```
dinhviethoa: Can you move `mCanCancel = true` after  loginIfNeeded() is done?

src/async/smtp/MCSMTPSendWithDataOperation.cpp
dinhviethoa: Maybe that's a good way to check if the cancel happened during connection after all ...

AlexKar: It's not clear why this additional check is required. In case if we call SMTPSendWithDataOperation::cancel before operation execution in serial queue, OperationQueue will filter this operation in OperationQueue::runOperations(), in other cases we have check in SMTPSession::internalSendMessage. Probably we can add additional check before loginIfNeeded(pError) to prevent unnecessary request.

src/core/smtp/MCSMTPSession.cpp
```diff
@@ -864,6 +914,23 @@ void SMTPSession::noop(ErrorCode * pError)
     }
 }
 
+void SMTPSession::cancelMessageSending()
+{
+    // main thread
+    if (mSmtp == NULL)
```
AlexKar: done

src/core/smtp/MCSMTPSession.cpp
AlexKar: done

dinhviethoa:
I updated binaries of libetpan and retrying travis.

src/core/smtp/MCSMTPSession.cpp
```diff
@@ -646,15 +683,25 @@ void SMTPSession::sendMessage(Address * from, Array * recipients, Data * message
     }
     MCLog("send");
     if ((mSmtp->esmtp & MAILSMTP_ESMTP_PIPELINING) != 0) {
-        r = mailesmtp_send_quit(mSmtp, MCUTF8(from->mailbox()), 0, NULL,
+        r = mailsmtp_data_message_quit_no_disconnect(mSmtp, MCUTF8(from->mailbox()), 0, NULL,
```
dinhviethoa: You probably wanted to use that function instead?
mailesmtp_send_quit_no_disconnect()

AlexKar: Thanks. I updated deps-versions.plist with correct version and added correct function to internalSendMessage. Currently it should build without any problems.

PR #1395, opened 2016-04-06
NNTP build fixes and IMAP runtime fixes

hhrowal:
Updated the NNTP interfaces to reflect changes made in the core. Fixed a few bugs that were causing crashes during message rendering and folderStatus operations.


dinhviethoa:
Love it!

PR #1399, opened 2016-04-15
Fix overwriting `pError` with `ErrorConnection` value

seaburg:
Hi!
Fix overwriting `pError` with `ErrorConnection` value after transition to `release` goto label

### DIFF
--- a/src/core/imap/MCIMAPSession.cpp
+++ b/src/core/imap/MCIMAPSession.cpp
@@ -3646,6 +3646,7 @@ void IMAPSession::storeFlagsAndCustomFlags(String * folder, bool identifier_is_u
             return;
         }
     }
+    * pError = ErrorNone;
 
     release:
     for(clistiter * iter = clist_begin(setList) ; iter != NULL ; iter = clist_next(iter)) {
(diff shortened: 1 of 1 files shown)


src/core/imap/MCIMAPSession.cpp
```diff
@@ -3743,6 +3743,7 @@ void IMAPSession::storeLabels(String * folder, bool identifier_is_uid, IndexSet
             return;
         }
     }
```
dinhviethoa: Could change the "return"s above to "goto release" too? It would be a great mem leak fix :)

src/core/imap/MCIMAPSession.cpp
```diff
@@ -3646,6 +3646,7 @@ void IMAPSession::storeFlagsAndCustomFlags(String * folder, bool identifier_is_u
...
+    * pError = ErrorNone;
```
dinhviethoa: Same here: Could change the "return"s above to "goto release" too? It would be a great mem leak fix :)

seaburg:
Fixed :)

dinhviethoa:
Thanks for fixing those!

PR #1400, opened 2016-04-15
More android fixes and additions

hhrowal:
Fixed the signature for a couple callback methods. Fixed local reference table overflow. Added missing methods to AbstractPart.

### DIFF
--- a/src/java/TypesUtils.cpp
+++ b/src/java/TypesUtils.cpp
@@ -348,16 +348,16 @@ static Object * hashmapJavaToObjectConverter(JNIEnv * env, jobject obj)
     jobject entrySet = (jobjectArray) env->CallObjectMethod(obj, method);
     javaClass = env->GetObjectClass(entrySet);
     method = env->GetMethodID(javaClass, "toArray", "()[Ljava/lang/Object;");
-    jobjectArray array = (jobjectArray) env->CallObjectMethod(obj, method);
+    jobjectArray array = (jobjectArray) env->CallObjectMethod(entrySet, method);
     int count = (int) env->GetArrayLength(array);
     for(int i = 0 ; i < count ; i ++) {
         jobject entry = env->GetObjectArrayElement(array, i);
         javaClass = env->GetObjectClass(entry);
         method = env->GetMethodID(javaClass, "getKey", "()Ljava/lang/Object;");
-        jobject key = env->CallObjectMethod(obj, method);
+        jobject key = env->CallObjectMethod(entry, method);
         Object * mcKey = javaToMCObject(env, key);
         method = env->GetMethodID(javaClass, "getValue", "()Ljava/lang/Object;");
-        jobject value = env->CallObjectMethod(obj, method);
+        jobject value = env->CallObjectMethod(entry, method);
         Object * mcValue = javaToMCObject(env, value);
         result->setObjectForKey(mcKey, mcValue);
     }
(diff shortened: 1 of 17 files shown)


src/java/com/libmailcore/HTMLRendererTemplateCallback.java
```diff
@@ -12,9 +12,9 @@
     /** Returns true if an attachment should (such as an image) be shown. */
     boolean shouldShowPart(AbstractPart part);
     /** Returns the values used in the header template for a given template. */
-    Map<String, String> templateValuesForHeader(MessageHeader header);
+    Map<String, Object> templateValuesForHeader(MessageHeader header);
```
dinhviethoa: Could you tell me why you changed it to Object? Values are always String.

src/java/TypesUtils.cpp
```diff
@@ -342,6 +342,7 @@ static bool isJavaMap(JNIEnv * env, jobject obj)
 
 static Object * hashmapJavaToObjectConverter(JNIEnv * env, jobject obj)
 {
+	env->PushLocalFrame(20);
```
dinhviethoa: Could you let me know why it's required?

dinhviethoa:
Some comments, everything else looks good!

src/java/TypesUtils.cpp
hhrowal: The local object references are usually cleaned up when control is returned to the JVM. However, during the htmlRendering operation, numerous calls are made through the JNI to gather the data necessary for rendering. Control is not returned to the JVM until all the callback methods have been executed. So much data is returned from the callbacks that he local object reference table fills up (512 entries) and causes app termination. This just releases the object references as soon as they are no longer required.

src/java/com/libmailcore/HTMLRendererTemplateCallback.java
hhrowal: The default callback in src/core/renderer/MCHTMLRendererCallback.cpp returns empty adds empty HashMaps to the result.

result->setObjectForKey(MCSTR("NOFROM"), mailcore::HashMap::hashMap());

In order to mimic that behavior, it was necessary to return a Map<String, Object>.

hhrowal: My comment was "edited". It is necessary to return `Map<String, Object> `

src/java/TypesUtils.cpp
dinhviethoa: What does 20 or 10 means? is it going to be the limit?
If that's just a hint and not the limit, could you use the same value everywhere (and maybe use a constant)?

src/java/TypesUtils.cpp
```diff
@@ -351,6 +352,7 @@ static Object * hashmapJavaToObjectConverter(JNIEnv * env, jobject obj)
     jobjectArray array = (jobjectArray) env->CallObjectMethod(entrySet, method);
     int count = (int) env->GetArrayLength(array);
     for(int i = 0 ; i < count ; i ++) {
+		env->PushLocalFrame(10);
```
dinhviethoa: Could you fix indentation here?

src/java/TypesUtils.cpp
dinhviethoa: There's also an indentation issue.

hhrowal: I see now there are other places in TypesUtils.cpp that already call PushLocalFrame with a value of 32. I will create a constant and make sure all calls use that constant

src/java/TypesUtils.cpp
hhrowal: Created LOCAL_FRAME_CAPACITY constant and fixed indentation.

dinhviethoa:
Thanks for helping make the Android support a real thing!

PR #1404, opened 2016-04-20
Try use IMAP connection with minimum queue size among connections tha…

disaykin:
…t already selected into needed folder

### DIFF
--- a/src/async/imap/MCIMAPAsyncSession.cpp
+++ b/src/async/imap/MCIMAPAsyncSession.cpp
@@ -286,11 +286,21 @@ IMAPAsyncConnection * IMAPAsyncSession::session()
 IMAPAsyncConnection * IMAPAsyncSession::sessionForFolder(String * folder, bool urgent)
 {
     if (folder == NULL) {
-        return availableSession();
+        return matchingSessionForFolder(NULL);
     }
     else {
         IMAPAsyncConnection * s = NULL;
+
+        // try find session with empty queue, selected to the folder
+        s = sessionWithMinQueue(true, folder);
+        if (s != NULL && s->operationsCount() == 0) {
+            s->setLastFolder(folder);
+            return s;
+        }
+
         if (urgent && mAllowsFolderConcurrentAccessEnabled) {
+            // in urgent mode try reuse any available session with
+            // empty queue or create new one, if maximum connections limit does not reached.
             s = availableSession();
             if (s->operationsCount() == 0) {
                 s->setLastFolder(folder);
@@ -298,6 +308,7 @@ IMAPAsyncConnection * IMAPAsyncSession::sessionForFolder(String * folder, bool u
             }
         }
 
+        // otherwise returns session with minimum size of queue among selected to the folder.
         s = matchingSessionForFolder(folder);
         s->setLastFolder(folder);
         return s;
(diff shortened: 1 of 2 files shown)


dinhviethoa:
There's usually only one connection that selected a given folder.
Could you let me know if you had situation or introduced situation where several connections have the same folder selected?
It might break some IMAP assumptions.
The only situation when I'm doing it is when the urgent flag is set up.

smirn0v:
We have proactive attachments prefetching. Situation with multiple connections with the same selected folder might be introduced with the following steps:

1. Folders list request
2. Attachment load for the message from INBOX folder (urgent = YES)

* At this stage mailcore2 will open 2 connections. One with no selected folder and one with 'INBOX' folder selected. *

3. UID FETCH for INBOX folder.

* At this stage mailcore2 will chose (1) (one with no folder selected) connection for the (3) request and as a result we will get 2 connections with INBOX folder selected. *

Subsequent request related to INBOX folder will chose only first connection, ignoring the fact that there is second one that has INBOX folder selected too.

With the change proposed by Dmitry – mailcore2 will evenly distribute subsequent operations between two connections.

Could you please elaborate what IMAP assumptions it may break and what solution you propose ?

disaykin:
Yes, we are using the urgent flag for some operations, typically for user triggered actions.

In case of active work with mails in one folder (user opens a message, then views its attachment, then cancels loading, then goes to the next message, then opens its attachment, then cancels loading, then returns to message list and pulls to refresh messages list) may occur situation with several connections that have the same folder selected. If first connection is currently used for downloading of big attachment, and second connection does not in use, it would be better use second connection for next request.

src/async/imap/MCIMAPAsyncSession.cpp
```diff
@@ -338,19 +347,37 @@ IMAPAsyncConnection * IMAPAsyncSession::availableSession()
...
-IMAPAsyncConnection * IMAPAsyncSession::matchingSessionForFolder(String * folder)
+IMAPAsyncConnection * IMAPAsyncSession::sessionForFolderWithMinQueue(String * folder)
 {
-    for(unsigned int i = 0 ; i < mSessions->count() ; i ++) {
-        IMAPAsyncConnection * currentSession = (IMAPAsyncConnection *) mSessions->objectAtIndex(i);
-        if (currentSession->lastFolder() != NULL) {
-            if (currentSession->lastFolder()->isEqual(folder)) {
-                return currentSession;
+    IMAPAsyncConnection * chosenSession = NULL;
+    unsigned int minOperationsCount = std::numeric_limits<unsigned int>::max();
```
dinhviethoa: Could you stay away from stdc++ as much as possible?
I'd rather use -1 as an indication that no value has been set yet.
(it implies casting thing but I'm fine with it)

dinhviethoa: Actually chosenSession being NULL is also a good indication (see availableSession).

dinhviethoa:
@smirn0v the assumption is that when mAllowsFolderConcurrentAccessEnabled is NO, we can't have two connections to the same folder. UW-IMAP won't work properly in that situation.
Anyway, it sounds like your use case looks good.

sessionForFolderWithMinQueue() and availableSession() look very similar.
Is there a way to refactor something here?

disaykin:
Done

dinhviethoa:
Sorry for the delay. I'll try to take a deep look. That's kind of critical part of the code. I'm trying to merge it with care.

disaykin:
Hi. Can you tell when you look at this patch? Thanks!

src/async/imap/MCIMAPAsyncSession.h
```diff
@@ -214,6 +214,7 @@ namespace mailcore {
 
         virtual IMAPAsyncConnection * session();
         virtual IMAPAsyncConnection * matchingSessionForFolder(String * folder);
+        virtual IMAPAsyncConnection * sessionWithMinQueue(bool filterByFolder = false, String * folder = NULL);
```
dinhviethoa: Could you remove the default values of the parameters?

src/async/imap/MCIMAPAsyncSession.cpp
```diff
@@ -306,58 +315,55 @@ IMAPAsyncConnection * IMAPAsyncSession::sessionForFolder(String * folder, bool u
...
+    IMAPAsyncConnection * chosenSession = NULL;
+    unsigned int minOperationsCount = UINT_MAX;
+
+    for (unsigned int i = 0 ; i < mSessions->count() ; i ++) {
+        IMAPAsyncConnection * s = (IMAPAsyncConnection *) mSessions->objectAtIndex(i);
+        if (s->operationsCount() < minOperationsCount) {
+            bool matched = true;
+            if (filterByFolder) {
+                matched = ((folder != NULL && s->lastFolder() != NULL && s->lastFolder()->isEqual(folder))
+                           || (folder == NULL && s->lastFolder() == NULL));
```
dinhviethoa: What would be the purpose of matching against folder == NULL?

src/async/imap/MCIMAPAsyncSession.cpp
```diff
@@ -306,58 +315,55 @@ IMAPAsyncConnection * IMAPAsyncSession::sessionForFolder(String * folder, bool u
...
+}
+
+IMAPAsyncConnection * IMAPAsyncSession::sessionWithMinQueue(bool filterByFolder, String * folder)
+{
+    IMAPAsyncConnection * chosenSession = NULL;
+    unsigned int minOperationsCount = UINT_MAX;
+
+    for (unsigned int i = 0 ; i < mSessions->count() ; i ++) {
+        IMAPAsyncConnection * s = (IMAPAsyncConnection *) mSessions->objectAtIndex(i);
+        if (s->operationsCount() < minOperationsCount) {
```
dinhviethoa: could you change it to if ((chosenSession == NULL) || (s->operationsCount() < minOperationsCount) )
then, initialize minOperationsCount to zero.

and remove `#include <limits.h>`

dinhviethoa:
I know there wasn't lots of good comments in the first place but
Could you add more comments to make it easier to understand?
Thanks!

src/async/imap/MCIMAPAsyncSession.cpp
disaykin: (folder == NULL) means that we must search connections that does not selected to any folder.

disaykin:
Done.

src/async/imap/MCIMAPAsyncSession.cpp
```diff
@@ -306,58 +317,63 @@ IMAPAsyncConnection * IMAPAsyncSession::sessionForFolder(String * folder, bool u
...
-            if (currentSession->lastFolder()->isEqual(folder)) {
-                return currentSession;
+    // try find session with minimum size of queue among selected to the folder.
+    IMAPAsyncConnection * chosenSelectedSession = sessionWithMinQueue(true, folder);
+    if (chosenSelectedSession != NULL) {
+        return chosenSelectedSession;
+    }
+
+    // try find session with minimum size of queue among non-selected to any folder ones.
+    IMAPAsyncConnection * chosenNonSelectedSession = sessionWithMinQueue(true, NULL);
```
dinhviethoa: I like to think that non-selected is equivalent to INBOX.
It may avoid creating a separate connections for requests without a specific folder.

disaykin: The main idea is in reusing of available session already selected to needed folder.
Since the SELECT command is very difficult for the server and can take a very long time, we have to minimize the number of current folder changes.

In case of folder == NULL in sessionForFolder() is used another branch of code:
    ```
if (folder == NULL) {
        return availableSession();
    }
```

dinhviethoa: Indeed but we could tight the "non-selected" sessions to the one with "INBOX" or NULL. It wouldn't involve additional SELECT commands.

disaykin: Do you propose to reuse session selected to INBOX for "non-selected" request?

dinhviethoa: That's it! :)

disaykin: But it is working as you want already. Method sessionForFolder() returns for folder == NULL the result of calling of availableSession(). Method availableSession returns a session with a minimum size of queue, creating a new session only if the session found has non-empty queue and if it have the opportunity to create a new connection.
Or am I wrong?

dinhviethoa: The issue is that it's potentially going to choose a folder that's not INBOX and bouncing around with other folders. I'd prefer to keep it tight to a specific folder: INBOX.

disaykin: Done

dinhviethoa:
Thanks! That's a good improvement!

PR #1406, opened 2016-04-22
Fix for imap servers which use non-standart name for inbox folder

disaykin:


src/core/imap/MCIMAPSession.cpp
```diff
@@ -1468,11 +1468,22 @@ Array * /* IMAPFolder */ IMAPSession::fetchAllFolders(ErrorCode * pError)
...
+                    break;
+                }
+            }
+
+            if (!hasInbox) {
+                r = mailimap_list(mImap, "", "INBOX", &imap_folders);
+                Array * inboxResult = resultsWithError(r, imap_folders, pError);
+                if (* pError == ErrorConnection || * pError == ErrorParse)
+                    mShouldDisconnect = true;
+                result->addObjectsFromArray(inboxResult);
```
dinhviethoa: Could you add an extra `hasInbox = true;` just to make it consistent?
Could you let me know which server has this issue?
Thanks!

disaykin:
Done.

PR #1411, opened 2016-04-27
Add settings for o2.pl and wp.pl mail providers

disaykin:


dinhviethoa:
For the other folders, could you add the ones of default? Thanks!

disaykin:
1) O2 provides folder flags, so no need in default folder names in providers.json.

2) It seems that WP.pl does not provide access to the drafts folder via IMAP.

resources/providers.json
```diff
@@ -1278,5 +1278,52 @@
...
+                     }
+                     ]
+        },
+        "domain-match": [
+                         "wp\\.pl"
+                         ],
+        "mailboxes": {
+            "spam": "SPAM",
+            "sentmail": "Wys&AUI-ane",
+            "trash": "Kosz"
```
dinhviethoa: Could you add those ones here?
```
            "drafts":"Drafts",
            "allmail":"Archive",
```

resources/providers.json
```diff
@@ -1278,5 +1278,52 @@
...
+                     {
+                     "port": 465,
+                     "hostname": "poczta.o2.pl",
+                     "ssl": true
+                     }
+                     ]
+        },
+        "domain-match": [
+                         "o2\\.pl"
+                         ]
```
dinhviethoa: Could add the mailboxes here anyway?
For consistency, I think there's the fallback for other one too.

disaykin:
Consistency is a good reason. Done.

PR #1417, opened 2016-05-05
Save SMTP server last response not only after message sending but aft…

disaykin:
…er login too

### DIFF
--- a/src/core/smtp/MCSMTPSession.cpp
+++ b/src/core/smtp/MCSMTPSession.cpp
@@ -258,6 +258,10 @@ void SMTPSession::unsetup()
 
 void SMTPSession::connectIfNeeded(ErrorCode * pError)
 {
+    // Clear out the last response from previous command.
+    MC_SAFE_RELEASE(mLastSMTPResponse);
+    mLastSMTPResponseCode = 0;
+
     if (mSmtp != NULL) {
         // In case pipelining is available, libetpan will disconnect the session.
         if (mSmtp->stream == NULL) {
@@ -287,6 +291,7 @@ void SMTPSession::connect(ErrorCode * pError)
         case ConnectionTypeStartTLS:
             MCLog("connect %s %u", MCUTF8(hostname()), (unsigned int) port());
             r = mailsmtp_socket_connect(mSmtp, MCUTF8(hostname()), port());
+            saveLastResponse();
             if (r != MAILSMTP_NO_ERROR) {
                 * pError = ErrorConnection;
                 goto close;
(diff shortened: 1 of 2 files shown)


dinhviethoa:
Could you clear our the last response in connectIfNeeded()?

disaykin:
For example, smtp.wp.pl server sometimes respond this way:

`dmitrijisajkin$ telnet smtp.wp.pl 465
Trying 212.77.101.1...
Connected to smtp.wp.pl.
Escape character is '^]'.
550 [IP:62.210.105.116] polaczenia z Twojego adresu IP sa zablokowane administracyjnie / mail from your IP address is administratively disabled
Connection closed by foreign host.`

For now, this situation treats by libetpan as SSL handshake failure. But in the future, somebody may improve this code, and libetpan will be return "550 [IP:62.210.105.116] polaczenia z Twojego adresu IP sa zablokowane administracyjnie / mail from your IP address is administratively disabled" as response string. This response may be shown to user as hint about problem resolution.

dinhviethoa:
I meant, when starting a new command, we need to clean up the old response from the previous command.
I think the best location would be in connectIfNeeded().

disaykin:
Done.

src/core/smtp/MCSMTPSession.cpp
```diff
@@ -717,17 +745,16 @@ void SMTPSession::internalSendMessage(Address * from, Array * recipients, Data *
         mailsmtp_quit(mSmtp);
     }
     esmtp_address_list_free(address_list);
+    saveLastResponse();
 
     String * response;
```
dinhviethoa: remove these 8 lines altogether. They don't seem to be useful.

disaykin:
Done

src/core/smtp/MCSMTPSession.cpp
```diff
@@ -717,25 +745,15 @@ void SMTPSession::internalSendMessage(Address * from, Array * recipients, Data *
...
+    saveLastResponse();
 
     if ((r == MAILSMTP_ERROR_STREAM) || (r == MAILSMTP_ERROR_CONNECTION_REFUSED)) {
         * pError = ErrorConnection;
         mShouldDisconnect = true;
         goto err;
     }
     else if (r == MAILSMTP_ERROR_EXCEED_STORAGE_ALLOCATION) {
-        if ((response != NULL) && (response->locationOfString(MCSTR("5.7.0")) != -1)) {
+        if ((mLastSMTPResponse != NULL) && (mLastSMTPResponse->locationOfString(MCSTR("5.7.0")) != -1)) {
```
dinhviethoa: ha, my bad, it was still used here! Anyway, it's good either way. I'm merging the PR.

dinhviethoa:
Thanks!

PR #1420, opened 2016-05-08
Prevent demo crash in _convertToJPEGData when image data cannot be converted

geoffcoward:
If a message contains images where the data is not a valid image, the demo crashes. These changes prevent the crash.

One source of such messages is from old Mac mail clients which include both image data and resource forks as two separate images in emails.

### DIFF
--- a/example/ios/iOS UI Test/iOS UI Test/MCTMsgViewController.mm
+++ b/example/ios/iOS UI Test/iOS UI Test/MCTMsgViewController.mm
@@ -254,38 +254,27 @@ typedef void (^DownloadCallback)(NSError * error);
     [info setObject:(id) [NSNumber numberWithFloat:(float) IMAGE_PREVIEW_WIDTH] forKey:(__bridge id) kCGImageSourceThumbnailMaxPixelSize];
     thumbnail = CGImageSourceCreateThumbnailAtIndex(imageSource, 0, (__bridge CFDictionaryRef) info);
 
-    CGImageDestinationRef destination;
-    NSMutableData * destData = [NSMutableData data];
-
-    if (destData != nil) {
+    if (thumbnail != nil) {
+        
+        CGImageDestinationRef destination;
+        NSMutableData * destData = [NSMutableData data];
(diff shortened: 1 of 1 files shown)


example/ios/iOS UI Test/iOS UI Test/MCTMsgViewController.mm
```diff
@@ -257,19 +257,35 @@ - (NSData *) _convertToJPEGData:(NSData *)data {
...
 
-    destination = CGImageDestinationCreateWithData((__bridge CFMutableDataRef) destData,
-                                                   (CFStringRef) @"public.jpeg",
-                                                   1, NULL);
-    
-    CGImageDestinationAddImage(destination, thumbnail, NULL);
-    CGImageDestinationFinalize(destination);
-
-    CFRelease(destination);
+    if (destData != nil) {
```
dinhviethoa: destData will always be != nil since it's been assigned above.

example/ios/iOS UI Test/iOS UI Test/MCTMsgViewController.mm
```diff
@@ -257,19 +257,35 @@ - (NSData *) _convertToJPEGData:(NSData *)data {
...
+        destination = CGImageDestinationCreateWithData((__bridge CFMutableDataRef) destData,
+                                                       (CFStringRef) @"public.jpeg",
+                                                       1, NULL);
+        if ((destination != NULL) && (thumbnail != NULL)) {
```
dinhviethoa: destination is likely to be always valid. I can't see a reason why it would be NULL.
(maybe memory exhaustion but terrible things would happen before then, we can just ignore that case)

dinhviethoa:
My guess is that you could just check thumbnail != NULL and only create the data in this case.

geoffcoward:
Updated the changes per your suggestions to only create data if "thumbnail" is created.

dinhviethoa:
Thanks!

geoffcoward:
Glad to contribute in some way!

PR #1423, opened 2016-05-10
Automaticaly enable Rambler workaround when we are downloading attac…

disaykin:
…ment with base64 or uuencode encoding

### DIFF
--- a/src/core/imap/MCIMAPSession.cpp
+++ b/src/core/imap/MCIMAPSession.cpp
@@ -2826,10 +2826,20 @@ Data * IMAPSession::fetchMessageAttachment(String * folder, bool identifier_is_u
     section = mailimap_section_new_part(section_part);
     fetch_att = mailimap_fetch_att_new_body_peek_section(section);
     fetch_type = mailimap_fetch_type_new_fetch_att(fetch_att);
-    
+
+#ifdef LIBETPAN_HAS_MAILIMAP_RAMBLER_WORKAROUND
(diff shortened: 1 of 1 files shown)


src/core/imap/MCIMAPSession.cpp
```diff
@@ -2826,10 +2826,20 @@ Data * IMAPSession::fetchMessageAttachment(String * folder, bool identifier_is_u
     section = mailimap_section_new_part(section_part);
     fetch_att = mailimap_fetch_att_new_body_peek_section(section);
     fetch_type = mailimap_fetch_type_new_fetch_att(fetch_att);
-    
+
+#ifdef LIBETPAN_HAS_MAILIMAP_RAMBLER_WORKAROUND
+    if ((encoding == EncodingBase64 || encoding == EncodingUUEncode) && (mHostname->compare(MCSTR("imap.rambler.ru")) == 0)) {
```
dinhviethoa: I merged it a little too soon.
All the workaround matching are usually here:
https://github.com/disaykin/mailcore2/blob/2f179e36b3289d056f432cc5854c9cd5489e5888/src/core/imap/MCIMAPSession.cpp#L644
Is there a way you could send a PR to make it consistent with other workarounds?

dinhviethoa: I mean setting a flag mRamblerRuServer = YES at this location and using it here.

disaykin: Done. But I detect Rambler server by hostname because it has non-unique welcome string "* OK IMAP4 ready".

PR #1424, opened 2016-05-11
Detect Rambler server on connect stage

disaykin:


src/core/imap/MCIMAPSession.cpp
```diff
@@ -652,6 +653,9 @@ void IMAPSession::connect(ErrorCode * pError)
             UNLOCK();
             mNamespaceEnabled = true;
         }
+#ifdef LIBETPAN_HAS_MAILIMAP_RAMBLER_WORKAROUND
```
dinhviethoa: You can remove the #ifdef here.

disaykin:
Done

dinhviethoa:
Thanks!

PR #1435, opened 2016-05-17
Fix permission access of source files that was executable

klefevre:
Some source files had wrong permission access .i.e. chmod 755 to 644.


dinhviethoa:
Thanks for fixing those!

PR #1436, opened 2016-05-18
Store last server response on login command in case of parse error

disaykin:


src/core/imap/MCIMAPSession.cpp
```diff
@@ -832,6 +832,15 @@ void IMAPSession::login(ErrorCode * pError)
     else if (r == MAILIMAP_ERROR_PARSE) {
         mShouldDisconnect = true;
         * pError = ErrorParse;
+
+        String * response = MCSTR("");
+        if (mImap->imap_response != NULL) {
+            response = String::stringWithUTF8Characters(mImap->imap_response);
```
dinhviethoa: Since it's a parsing error, that one will be empty.

src/core/imap/MCIMAPSession.cpp
```diff
@@ -832,6 +832,15 @@ void IMAPSession::login(ErrorCode * pError)
...
+        } else if (mImap->imap_stream_buffer != NULL) {
+            response = String::stringWithUTF8Characters(mImap->imap_stream_buffer->str);
```
dinhviethoa: I think we should create a separate accessor for that buffer and keep is as raw Data * since we're not guaranteed that it will convert smoothly as UTF8.

dinhviethoa: The accessor could be name "protocolData". And maybe document that it will be filled only only when the IMAP parser will fail (on LOGIN for now).

disaykin:
Done

src/core/imap/MCIMAPSession.cpp
```diff
@@ -355,6 +355,7 @@ void IMAPSession::init()
     mAutomaticConfigurationDone = false;
     mShouldDisconnect = false;
     mLoginResponse = NULL;
+    mLoginUnparsedResponseData = NULL;
```
dinhviethoa: Could you change the name to mUnparsedResponseData since we might also apply to other areas than just login?

FETCH response can fail to be parsed too on some broken servers :(

src/core/imap/MCIMAPSession.cpp
```diff
@@ -832,6 +842,13 @@ void IMAPSession::login(ErrorCode * pError)
     else if (r == MAILIMAP_ERROR_PARSE) {
         mShouldDisconnect = true;
         * pError = ErrorParse;
+
+        Data * unparsed_response = Data::data();
+        if (mImap->imap_stream_buffer != NULL) {
+            unparsed_response = Data::dataWithBytes(mImap->imap_stream_buffer->str, (unsigned int) mImap->imap_stream_buffer->len);
+        }
+        MC_SAFE_REPLACE_COPY(Data, mLoginUnparsedResponseData, unparsed_response);
```
dinhviethoa: Use `MC_SAFE_REPLACE_RETAIN()` To avoid an extra copy.

disaykin:
Done

PR #1439, opened 2016-05-24
Additional providers search options

NIkita-Ruzanov:
Additional providers search options that can be used instead of regexp for faster search in large providers jsons

### DIFF
--- a/src/core/provider/MCMailProvider.cpp
+++ b/src/core/provider/MCMailProvider.cpp
@@ -30,6 +30,9 @@ void MailProvider::init()
     mSmtpServices = new Array();
     mPopServices = new Array();
     mDomainMatch = new Array();
+    mDomainMatchNoRegexp = new Array();
+    mDomainEndsWith = new Array();
+    mDomainStartsWith = new Array();
     mDomainExclude = new Array();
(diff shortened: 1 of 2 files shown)


dinhviethoa:
Hey Nikita, I'm not sure to understand the pull request here since everything is covered with domain-match and regexp. Could you give more details about your throughts?
There's also probably some ways to cache the result of regcomp() to improve speed.

NIkita-Ruzanov:
Hi, indeed all main search functionality is covered by domain-match and regexp. Main point of my pull request is to get rid of necessity to use regexp for domain search. After we expanded our providers.json to include around 200 entries search time changed from 2-3ms per provider to 15-30ms. So we decided to change to simple string search as it gives us search time of 0.3-0.7ms per provider and most providers don't require any complicated domain search which may justify use of regexp. This is crucial to our application as we are performing search at start up of application.

dinhviethoa:
Could you provide this providers.json file (through a pull request)?
1/ It would be a good contribution to the project.
2/ It would help testing and improving that area of the code.
Thanks!

lksnmnn:
Yeah. Please share the providers.json :)

PR #1446, opened 2016-06-05
add Objective-C generics

larryonoff:


larryonoff:
I'm unable to solve ```error: type arguments cannot be applied to non-parameterized class``` error yet.

dinhviethoa:
Do you know if that would break build of projects using the API without generics?

larryonoff:
@dinhviethoa good catch!

I've checked how this case is handled by others, so for example Parse, https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/pull/190/files#diff-a3220900c6310732ec7a2fd8a7ff8fa0R440

```objc
#if __has_feature(objc_generics) || __has_extension(objc_generics)
#  define PF_GENERIC(type) <type>
#else
#  define PF_GENERIC(type)
#  define PFGenericObject PFObject *
#endif
```

so we can introduce macro like
```objc
#if __has_feature(objc_generics) || __has_extension(objc_generics)
#  define MCO_GENERIC(type) <type>
#  define MCO_GENERIC(type, type) <type, type>
#else
#  define MCO_GENERIC(type)
#  define MCO_GENERIC(type, type)
#endif
```

What file should I add this macro?

larryonoff:
@dinhviethoa some other examples
- https://gist.github.com/JaviSoto/e75a2ba60f0aad1721b4
- https://gist.github.com/smileyborg/d513754bc1cf41678054

dinhviethoa:
What I meant is rather people having a compiler that supports generics but uses the old API definition without generics in their code. Do you know if their build will fail if we add generics in the headers?

larryonoff:
@dinhviethoa I'm not sure, but I assume that their build will fail (if Xcode < 7) according to this url, see https://twitter.com/javi/status/608395796288012288. So the macro above allows conditionally use ObjC generics when available.

dinhviethoa:
1/ If Xcode >= 7 and people don't update how they use the framework, is it going to be smooth?
2/ On Travis, it's weird that the build fails only on cmake though. Do you have any idea why?

I'm fine with assuming every one upgraded to Xcode >= 7.

larryonoff:
> If Xcode >= 7 and people don't update how they use the framework, is it going to be smooth?

I'm not sure that I understand this question, but please see my commends below.
Xcode >= 7 and someone migrates into the new framework version (with generics).
 - if use framework in Obj-C code -> nothing needs to be changed.
 - if use framework in Swift code -> code needs to be updated, but generally just removing unnecessary code (e.g. casts from ```AnyObject``` into appropriate type). This's an often case for Swift developers at the moment, e.g. because new versions of Swift often have breaking changes.

> On Travis, it's weird that the build fails only on cmake though. Do you have any idea why?

I don't know why, but I'm trying to find the solution. It only fails for OS X.

> I'm fine with assuming every one upgraded to Xcode >= 7.

This is good, because we don't need to introduce defines in this case.

larryonoff:
How do you feel about the next steps below.
1. Fix Travis build.
2. Review the code and find the places where the generics were missed.

dinhviethoa:
Sounds good to me.

do you need to use `<__kindof MCOAddress *>` instead of just `<MCOAddress *>`?

larryonoff:
> do you need to use <__kindof MCOAddress *> instead of just <MCOAddress *>?

```__kindof``` is convenient way to tell the compiler that e.g. ```NSArray``` contains objects that are inherited from concrete type.

Apple documentation says the following:
> KindOf. Objects declared as __kindof types express “some kind of X” to the compiler and can be used within generic parameters to constrain types to a particular class or its subclasses. Using __kindof allows constraints to be more flexible than an explicit class, and more explicit than just using id.

For a better explanation please see http://stackoverflow.com/a/31399395/4001725

I've reviewed the code and removed the places where I assume ```__kindof``` isn't needed. But there's some places where it remains.

**Comment by larryonoff** [comment]
@dinhviethoa could you please check the following files where generics can be added, because I'm not sure what types can be used here.
- _imap/MCOIMAPCopyMessagesOperation.h_

```NSDictionary * __nullable uidMapping```
```
- (void) start:(void (^)(NSError * __nullable error, NSDictionary * __nullable uidMapping))completionBlock;
```
- _imap/MCOIMAPIdentityOperation.h_

```NSDictionary * __nullable serverIdentity```
```
- (void) start:(void (^)(NSError * __nullable error, NSDictionary * __nullable serverIdentity))completionBlock;
```

dinhviethoa:
NSDictionary * __nullable uidMapping: NSNumber, NSNumber

and for this one, I noticed that there was a type mismatch between C++ and Obj-C. I need to fix that first.
NSDictionary * __nullable serverIdentity.
Maybe assume for now that it's NSString, NSString.

larryonoff:
I've updated dictionaries above.

dinhviethoa:
About kindof. I think you should use it only when there are abstract classes.
The names of the classes should show "abstract".

MCOAddress is not abstract.

larryonoff:
> About kindof. I think you should use it only when there are abstract classes.

I agree. I've removed ```kindof``` for not abstract classes.

larryonoff:
@dinhviethoa as a short status. I haven't been able to fix cmake build yet, but I'll proceed again as soon as I have more free time.

dinhviethoa:
Any updates on it?

larryonoff:
@dinhviethoa nope. I don't have time to work on the issue, but I'll proceed as soon as I have time.

PR #1460, opened 2016-06-24
Compile mailcore2 locally via cocoapods

virtru-bjost:
This will allow users to compile mailcore2 locally via cocoapods so that they can specify specific revisions they want without the developers having to make builds.

add

`pod 'mailcore2-ios', :git => 'http://github.com/MailCore/mailcore2.git'` to your podfile and you can also specify specific commits to build from

### DIFF
--- a/mailcore2-ios.podspec.json
+++ b/mailcore2-ios.podspec.json
@@ -1,0 +1,26 @@
+{
+  "name": "mailcore2-ios",
+  "version": "0.6.4",
+  "summary": "Mailcore 2 for iOS",
+  "description": "MailCore 2 provide a simple and asynchronous API to work with e-mail protocols IMAP, POP and SMTP.",
+  "homepage": "http://libmailcore.com",
+  "license": {
+    "type": "BSD",
+    "file": "LICENSE"
+  },
+  "authors": "MailCore Authors",
+  "source": {
+    "http": "http://d.etpan.org/mailcore2-deps/mailcore2-ios/mailcore2-ios-12.zip",
+    "flatten": true
+  },
+  "platforms": {
+    "ios": 8.0
+  },
+  "header_dir": "MailCore",
+  "requires_arc": false,
+  "public_header_files": "include/MailCore/*.h",
+  "preserve_paths": "include/MailCore/*.h",
+  "vendored_libraries": "lib/libMailCore-ios.a",
+  "libraries": ["xml2", "iconv", "z", "c++", "resolv"],
+  "prepare_command": "curl -O https://github.com/MailCore/mailcore2/raw/master/LICENSE"
+}
--- a/mailcore2-ios.podspec.json
+++ b/mailcore2-ios.podspec.json
@@ -10,17 +10,17 @@
   },
   "authors": "MailCore Authors",
   "source": {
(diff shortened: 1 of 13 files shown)


mailcore2-ios.podspec.json
```diff
@@ -0,0 +1,26 @@
...
+  "source": {
+    "git": "http://github.com/MailCore/mailcore2.git",
+    "branch": "master"
+  },
+  "platforms": {
+    "ios": 8.0
+  },
+  "header_dir": "MailCore",
+  "requires_arc": false,
+  "public_header_files": "built/include/MailCore/*.h",
```
dinhviethoa: Could you use "cocoapods-build" as the name of the main folder for the build?
And add it to .gitignore?

scripts/build-mailcore2-ios.sh
```diff
@@ -6,13 +6,12 @@ popd > /dev/null
 
 . "$scriptpath/include.sh/build-dep.sh"
 
-url="https://github.com/mailcore/mailcore2.git"
```
dinhviethoa: Could you rather add a parameter to the script to choose a commit or to build master directly?
I'd like to keep a default commit number in that script.

scripts/deps-versions.plist
```diff
@@ -13,7 +13,7 @@
 	<key>mailcore2-framework-osx</key>
 	<string>9</string>
 	<key>mailcore2-ios</key>
-	<string>12</string>
+	<string>13</string>
```
dinhviethoa: Please revert that change.

scripts/include.sh/build-dep.sh
```diff
@@ -191,33 +171,9 @@ build_git_osx()
   mkdir -p "$srcdir"
 
   pushd . >/dev/null
-  mkdir -p "$builddir/downloads"
```
dinhviethoa: I'd like to avoid this change or maybe could you add a parameter to that bash function to tweak the the behavior without breaking the rest?

dinhviethoa:
Thanks for the pull request!
Could you make the changes to avoid breaking the existing behavior?

virtru-bjost:
@dinhviethoa I would make a copy of the original scripts and name them `-manual.sh` and use those for manual builds. I can make the other changes though.

dinhviethoa:
@virtru-bjost sounds good!

virtru-bjost:
@dinhviethoa changes pending check. Once / and if, you merge this in, you could also update the Cocoapods/Spec repo to use these new pod files. This will allow people doing first time mailcore installs to use the latest from master. Or you can make a tag and change the podspec.json to use that tag by default instead.

dinhviethoa:
I'm sorry, I think I misunderstood the build-...-manual.sh story.
Could you keep the original name of the scripts with the same behavior?
And have the duplicate of the scripts named build-...-cocoapods.sh?
It would be explicit enough and the patch review will be smoother.

Thanks!

virtru-bjost:
ok @dinhviethoa think this is what you were looking for.

dinhviethoa:
Thanks! I think it will be really helpful for the users of Cocoapods!

PR #1464, opened 2016-07-04
MCSMTPSession.cpp: outlook bcc fix

serjepatoff:
I've discovered that outlook.com requires presence of "bcc" field in rfc822 data blob. Otherwise message won't be delivered to bcc recipients without any notice from server.

### DIFF
--- a/src/core/smtp/MCSMTPSession.cpp
+++ b/src/core/smtp/MCSMTPSession.cpp
@@ -709,7 +709,7 @@ void SMTPSession::internalSendMessage(Address * from, Array * recipients, Data *
         return;
     }
     
-    if (!this->mOutlook) {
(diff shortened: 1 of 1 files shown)


dinhviethoa:
Could you confirm that the message is received by the Bcc recipient and the recipient without the Bcc field?

serjepatoff:
I tested this issue with 2 recipients in "RCPT TO:" SMTP command, and the same 2 recipients filled in rfc822 data (first in "To:" field, second in "Bcc:" field).

First run, without my pull request. Message was delivered to "To:" recipient only.
Second run, with my change. Message was delivered to both recipients.

After that I've checked up the second message on both recepients' sides, and "Bcc:" field was wiped out by outlook.com.

PR #1465, opened 2016-07-05
fix missing search by read/unread flag

raduraducu:


src/core/imap/MCIMAPSession.cpp
```diff
@@ -2760,6 +2760,16 @@ IndexSet * IMAPSession::search(String * folder, IMAPSearchKind kind, String * se
             expr = IMAPSearchExpression::searchContent(searchString);
             break;
         }
+        case IMAPSearchKindRead:
```
dinhviethoa: If you're adding those, could you also add the other search types that are flags-like?
- flagged
- answered
- draft
- deleted
- spam

dinhviethoa:
Thanks!

PR #1470, opened 2016-07-07
MCSMTPSession.cpp: outlook bcc fix (continued)

serjepatoff:
My apologizes for undiscovered bug.
Actual assignment to mOutlookServer (based on host name) occured after making decision on bcc wiping. When default value of mOutlookServer was true I didn't notice this misbehaviour.

### DIFF
--- a/src/core/smtp/MCSMTPSession.cpp
+++ b/src/core/smtp/MCSMTPSession.cpp
@@ -709,7 +709,7 @@ void SMTPSession::internalSendMessage(Address * from, Array * recipients, Data *
         return;
     }
     
-    if (!this->mOutlook) {
+    if (!this->mOutlookServer) {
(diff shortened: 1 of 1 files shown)


src/core/smtp/MCSMTPSession.cpp
```diff
@@ -77,6 +77,12 @@ SMTPSession::~SMTPSession()
 void SMTPSession::setHostname(String * hostname)
 {
     MC_SAFE_REPLACE_COPY(String, mHostname, hostname);
+    
+    if (hostname != NULL) {
```
dinhviethoa: Could you reset the value of mOutlookServer to false?

serjepatoff: For sure.

PR #1509, opened 2016-09-12
MCIMAPCheckAccountOperation: connectIfNeeded() then login() sequence …

serjepatoff:
It's dangerous to call connectIfNeeded(&error) and then login(&error). If we are already connected and mShouldDisconnect flag is set, then connectIfNeeded() will disconnect and then try to connect again. BUT. The last call to connect can fail, and after that when we try to call login() we'll catch MCAssert(mState == STATE_CONNECTED). Kaboom.

PS. It's rather hard to reproduce this error in real life, one should have unstable error-prone internet connection and issue a reconnect. But we have ~5k remote crash dumps that led us exactly to this place in code.


dinhviethoa:
Thanks!

PR #1517, opened 2016-09-26
adding smtp error: 535 5.7.0

andreapz:
authentication rejected: your password is too simple


dinhviethoa:
Is that error from a specific provider?
If it's the case, please prefix the enum.

andreapz:
is an error from mail.tiscali.it server

PR #1520, opened 2016-09-29
Smtp error: 535 5.7.0 on Tiscali smtp server

andreapz:


andreapz:
This pull request is made to fix add an error that came from tiscali smtp server when some contact try to send mail but have a simple password like 123456.

dinhviethoa:
Thanks!

Issue #1426, opened 2016-05-12
Using the new type attributions from Objective-C


CTMacUser:
To improve compatibility with Swift and its Swift-to-ObjC converter, Objective-C supports adding attributes to types and types within function descriptions that indicate is NIL parameters and/or returns are supported. And there are new attributes for the collections types (NSArray, NSSet, NSDictionary) to indicate their element types. Before trying to add them (and submit a pull request), I wonder if there are compatibility concerns to adding them.

dinhviethoa:
My only concern may be how much it may make the code unreadable.
Maybe you could show me the result on one of the classes?

klefevre:
@dinhviethoa +1 on this. This would improve a lot the swift compatibility. I understand your concern about making the code more readable but I think this is a mandatory task since Swift becomes the main language of Apple.

I made a gist where you can see how much a file should be modified : https://gist.github.com/klefevre/77045cebb53f8af854e553e6341fce3a

dinhviethoa:
It looks like there's going to be couple of back and forth about APIs.
We could start on some classes, do it progressively.

(Don't send a big pull request)

yasinbor:
I am jumping in but i have to ask, do you guys thinking about making swift example project any time soon?
I am exhausted to trying to read objective-c code.

PR #1530, opened 2016-10-13
Type annotations

Palleas:
I've started a Swift project using MailCore and I figured it would be nice to have more types. 
As mentioned in #1426, a gigantic pull request would be unreadable so I'm starting with what I'm
actually using and I'll open small PRs like this one if that works for you.

I can also add nullability annotations but I'm not sure what can be nil and what can't. Any idea on 
how I could get this information ? We can also assume that (almost) everything is nullable, like
the display name of an `MCOAddress` ?

Please let me know what you think, I'm looking forward to contributing to this life-saving project 😄


dinhviethoa:
Let's focus on generics for now.
Nullability is probably more difficult to address.

Palleas:
> Let's focus on generics for now.

Works for me!

dinhviethoa:
Please resend the pull request when the switch to xcodebuild is ready!

PR #1531, opened 2016-10-14
Add type annotations to some operations and MCPIMAPSession

Palleas:
Same as #1530 ! :shipit: 

See #1426.

### DIFF
--- a/src/objc/imap/MCOIMAPFetchFoldersOperation.h
+++ b/src/objc/imap/MCOIMAPFetchFoldersOperation.h
@@ -13,6 +13,7 @@
 /** This class implements an operation to fetch a list of folders. */
 
 #import <MailCore/MCOIMAPBaseOperation.h>
+@class MCOIMAPFolder;
 
 NS_ASSUME_NONNULL_BEGIN
 @interface MCOIMAPFetchFoldersOperation : MCOIMAPBaseOperation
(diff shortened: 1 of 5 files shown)


dinhviethoa:
It caused a build failure :(
https://travis-ci.org/MailCore/mailcore2/jobs/167513546

Palleas:
It seems to be an issue related to C++..?

Palleas:
It seems to be this part that fails on OSX. Do you want to restore `master` to a working state while I investigate ? It looks like type annotations doesn't play well with C++..? Could be a compiler version issue ? 🤔

dinhviethoa:
I'I going to revert that commit. Did it at least work locally for you?

Palleas:
Yeah it works in Xcode (8) but the issue happens in the CMake build and that I'm unable to run locally: https://github.com/MailCore/mailcore2/blob/master/scripts/travis/script.sh#L43-L49

dinhviethoa:
Is it building for i386? Maybe it may be why it's failing?

Palleas:
I'll try to build it locally but I miss some dependencies like `libetpan` so it may take some time. 

I also noticed that the build script still uses xctool and the travis build targets xcode 7. Is it ok if I open a PR for migrating to xcodebuild (or facebook new toy, xcbuild)?

dinhviethoa:
Sure. You can open a PR to use xcodebuild or xcbuild.
xcbuild would probably be preferred since it's probably less verbose than xcodebuild.

dinhviethoa:
I'm about to revert that commit. Could you give it a try to xcodebuild if it help?

Palleas:
Instead of xcbuild you mean?

dinhviethoa:
Yes. Then, switch to xcbuild later.

Palleas:
I'll work on that then!

Palleas:
(Just to be clear I'm still working on adding xcbuild as well but I haven't had much time lately 😊 )

dinhviethoa:
Please resend the pull request when the switch to xcodebuild is ready!

Issue #1537, opened 2016-10-19
Cocoapods trying to use xctool


jostster:
The cocoapods deps script is trying to use xctool which is no longer allowed with xcode 8.

PR #1538, opened 2016-10-19
Fix for issue #1537

jostster:
This fixes the dependency on xctool now that it isn't allowed in xcode 8


jostster:
Fix for issue #1537

dinhviethoa:
Thanks!

Issue #1479, opened 2016-07-15
Invalid Symlink - Your package contains a symbolic link 


atljeremy:
I just switched to using Carthage to manage my apps dependencies and got this error when I uploaded the latest version to iTunes Connect. Any ideas on how to address this?

```
Invalid Symlink - Your package contains a symbolic link 'Frameworks/MailCore.framework/Headers/Headers' which resolves to a location 'Frameworks/MailCore.framework/Headers/Versions/Current/Headers' that does not exist or is outside of the package.
```

Thanks!

atljeremy:
I found this (http://stackoverflow.com/a/36953033), which works, but it's a pain to have to open the archive, open the app, remove the symlink, then upload.

dinhviethoa:
Could you investigate with Carthage tool to see what's the issue here? (File an issue on their github)

atljeremy:
@dinhviethoa I've investigated this, and it has nothing to do with Carthage. It's due to the custom shell scripts you have in the `mailcore2.xcodeproj`. Here are the relevant lines that are creating the symlink to `Headers`. This needs to be removed as submissions to the App Store with this symlink will be rejected.

https://github.com/MailCore/mailcore2/blob/master/build-mac/mailcore2.xcodeproj/project.pbxproj#L3472

https://github.com/MailCore/mailcore2/blob/master/build-mac/mailcore2.xcodeproj/project.pbxproj#L3511

In those two libs you'll find this script:

```bash
rsync -a  \"$BUILT_PRODUCTS_DIR/include/MailCore/\" \"$BUILT_PRODUCTS_DIR/$PUBLIC_HEADERS_FOLDER_PATH\"
cd \"$BUILT_PRODUCTS_DIR/$FULL_PRODUCT_NAME\"
rm -f Headers
ln -sf \"Versions/Current/Headers\" Headers";
```

Notice the very last line in that script is creating the symlink `ln -sf \"Versions/Current/Headers\" Headers"`. That line should be removed.

Hope that helps clear things up.

atljeremy:
@dinhviethoa Why was this closed without any comments? Are you planning to remove this unnecessary symlink that is actually causing submissions to the app store to be rejected? I'm currently having to manually remove this symlink from the binary every time I want to upload a build to Apple.

otaviolima:
@atljeremy 

Faced the same issue. There are couple of options to solve the issue:

[...]

PR #1542, opened 2016-10-24
Remove symlink creation for osx and ios targets

otaviolima:
This symlink causes any app containing the generated MailCore2.Framework to be rejected by iTunes Connect.

Resolves: #1479


dinhviethoa:
Thanks!

Palleas:
It works the build if you use Carthage 😢

Palleas:
Fixed in #1544 😄

PR #1557, opened 2016-11-06
No longer remove the Headers folder

Palleas:
Following #1542 I was unable to use the framework using Carthage, when I looked at the built `framework`, I noticed that all the header files were gone. Do we really need to remove the `Headers` folder?

### DIFF
--- a/build-mac/mailcore2.xcodeproj/project.pbxproj
+++ b/build-mac/mailcore2.xcodeproj/project.pbxproj
@@ -3469,7 +3469,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


Palleas:
I'm still having having issues on the mac version though 🤔

otaviolima:
@Palleas I get the point of that they are breaking the Carthage build but when I made this PR I had this #1479 in mind. 
Are you sure you can upload a app to iTunes connect with the generated .framework?

Palleas:
@otaviolima I don't see why not because it's no longer a symlink but I only submitted a build to testflight for now...

otaviolima:
@Palleas Yes, paid better attention to the changes now. Somehow in my mind this commit was a revert of my previous commit.
Nevermind then, everything should be ok for AppStore submissions. :D

PR #1601, opened 2017-01-24
Add missing public headers to umbrella header

conradev:
Let me know if I should make any changes!

This fixes Xcode warnings about missing headers when using the framework.


conradev:
Investigating the build failure…

src/async/imap/MCAsyncIMAP.h
```diff
@@ -12,6 +12,8 @@
 
 #include <MailCore/MCIMAPAsyncSession.h>
 #include <MailCore/MCIMAPOperation.h>
+#include <MailCore/MCIMAPNoopOperation.h>
```
dinhviethoa: Why do you think it should be public?

dinhviethoa:
The build failure is due to the cmake build workflow.

conradev:
Hi Hoà!

The issue is that these headers are included in the built framework's `Headers` directory, but they are not referenced by the umbrella header. This causes a warning in Xcode. Perhaps the correct solution would be to make it so that these files don't end up in the `Headers` directory? I can make that change.

dinhviethoa:
Thanks!

PR #1619, opened 2017-03-10
Remove crypto & ssl libraries in podspec for osx

nikolay-kapustin:
For os x development these libs don't need by default.


petangi68:
when do you think this will make an official pod release?

PR #1680, opened 2017-10-30
Fix broken blockElements lock

bengotow:
Hey! This should be a static MC_LOCK_TYPE like the rest in this file.


dinhviethoa:
wow!

PR #1690, opened 2017-11-20
Fix calendar attachment appearing twice in the HTML output

fastred:
Bug:
Calendar attachment was appearing twice in the HTML body.

Reason:
If `preferredAlternative` was equal to `calendar`, we were adding output of `htmlForAbstractPart()` for the same `AbstractPart` twice to the `result` variable.


dinhviethoa:
Thanks for fixing it!

PR #1707, opened 2018-02-13
Fix crash when connection's owner is NULL

fastred:
H!. I've been observing a crash caused by `mConnection->owner()` being `NULL` in `queueStartRunning()`. I wasn't able to reproduce or even isolate it, though. It happens sporadically, in about 0.1% of all sessions.

We create a new session (and deallocate the old one) when a reachability status changes. My understanding is that in this case `IMAPAsyncSession` instance gets deallocated but `IMAPAsyncConnection` is still alive. Should this situation be even possible from the memory-management POV? I see that `IMAPAsyncConnection` retains `mOwner`, so I would assume that `IMAPAsyncSession` instance would always outlive the connection. Is there a better fix than what I propose?

Stack trace from Crashlytics:
screen shot 2018-02-13 at 13 05 11

### DIFF
--- a/src/async/imap/MCIMAPAsyncConnection.cpp
+++ b/src/async/imap/MCIMAPAsyncConnection.cpp
@@ -57,14 +57,18 @@ namespace mailcore {
 
         virtual void queueStartRunning() {
             mConnection->setQueueRunning(true);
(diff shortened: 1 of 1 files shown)


dinhviethoa:
Let's use this mitigation for now.

PR #1729, opened 2018-04-27
Fix incorrect russian subject in test data

yuryybk:
To pass this test for Android please rebuild libeptan for Android with iconv.
Libetpan update for Android with iconv was merged recently (https://github.com/dinhviethoa/libetpan/commit/3f77ed4d976c88405253a6f6a988a6e132a20a90)

Also Android mailcore requires addition version 7 of libetpan here (http://d.etpan.org/mailcore2-deps/libetpan-android/) and version changes for Android here (https://github.com/MailCore/mailcore2/blob/master/build-android/build.sh#L8)


dinhviethoa:
I need to investigate more why the test is not passing on Mac when applying this patch.

Saravia83c:
- [__]()

Issue #1730, opened 2018-05-04
How to create MCOAccountValidator?


diejmon:
MCOAccountValidator has no public available inits. MCOAccountValidator is subclass of MCOOperation which marks default init as NS_UNAVAILABLE.

PR #1736, opened 2018-05-14
Make MCOAccountValidator init method available

diejmon:


diejmon:
A fix for #1730

dinhviethoa:
It looks like the right fix. Thanks!

PR #1749, opened 2018-06-08
feat(IMAPSession): add support for moveMessagesOpeation to java library

farnabaz:
Right now java library does not have move operation, but this functionality already implemented in MailCore


dinhviethoa:
Thank you!

PR #1750, opened 2018-06-08
Add uidplus, acl, enable capabilities parsing

diejmon:


dbshdgw:
>  thank you for the quote

Issue #1482, opened 2016-07-20
Wrong JNI type in getField: Not Long but Java Object


vonox7:
When setting a ConnectionLogger or OperationQueueListener to SMTPSession the setupNative function fails with the following error:
`JNI GetObjectField called with pending exception java.lang.NoSuchFieldError: no "J" field "connectionLogger" in class "Lcom/libmailcore/SMTPSession;" or its superclasses`

The _connectionLogger/operationQueueListener_ field in SMTPSession is not a Long ("J") but an Object ("L..."), therefore the JNI GetFieldID function fails.

In File [JavaHandle.cpp](https://github.com/MailCore/mailcore2/blob/master/src/java/JavaHandle.cpp) the function getField fetches through JNI an object, but GetFieldID falsly assumes that the type to search is a Long instead of an Java Object:
```c
static jfieldID getField(JNIEnv * env, jobject obj, const char * fieldName)
{
    jclass c = env->GetObjectClass(obj);
    return env->GetFieldID(c, fieldName, "J");
}
```

The getField function is only called by getObjectField():
```c
jobject mailcore::getObjectField(JNIEnv *env, jobject obj, const char * fieldName)
{
    return env->GetObjectField(obj, getField(env, obj, fieldName));
}
```

Expected code: 
jclass c = env->GetFieldID(c, fieldName, **"Lcom/libmailcore/ConnectionLogger;"**)
and
jclass c = env->GetFieldID(c, fieldName, **"Lcom/libmailcore/OperationQueueListener;"**)

To solve this issue for all functions that use the getObjectField function, we would need an additional parameter that specifies the Java class we are looking for. (ConnectionLogger, OperationQueueListener...).

sasvarizoli:
+1

dinhviethoa:
Could you send a pull request to fix it?
Thanks!

vonox7:
I tried to fix it, but apparently there are more bugs related to that issue.

dinhviethoa:
@vonox7, would the following work?

```
static jfieldID getFieldForObject(JNIEnv * env, jobject obj, const char * fieldName)
{
    jclass c = env->GetObjectClass(obj);
    return env->GetFieldID(c, fieldName, "Ljava/lang/Object;");
}

jobject mailcore::getObjectField(JNIEnv *env, jobject obj, const char * fieldName)
{
    return env->GetObjectField(obj, getFieldForObject(env, obj, fieldName));
}
```

[...]

PR #1770, opened 2018-07-24
fix some wrong JNI signatures

wei-gong:


dinhviethoa:
Thanks for fixing it!

vonox7:
@wei-gong do you think this solves also #1482 ?

PR #1804, opened 2019-03-21
Fix few memory leaks

diejmon:


src/core/provider/MCMailProvidersManager.cpp
```diff
@@ -21,6 +21,10 @@ MailProvidersManager::MailProvidersManager() {
     init();
 }
 
+MailProvidersManager::~MailProvidersManager() {
+    delete mProviders;
```
dinhviethoa: Could you use the following instead?

```
MC_SAFE_RELEASE(mProvider);
```

diejmon: Done

dinhviethoa:
Thanks!

PR #1852, opened 2019-12-12
Fix superclass for MCONNTPDisconnectOperation

bbarenblat:
If `MCONNTPDisconnectOperation` inherits from `NSObject`, `MCONNTPDisconnectOperation+mco_objectWithMCObject:` will call a method that doesn’t exist (`-initWithMCOperation:`). Set it to extend MCONNTPOperation instead.


dinhviethoa:
Thank you!

PR #1855, opened 2020-01-03
Exclude DriverKit SDK from macOS SDK version variable.

futureimperfect:
The DriverKit SDK can appear in the output from `xcodebuild -showsdks`, and as such `macosx19.0` is parsed as a macOS SDK in build scripts, causing them to fail. This change explicitly excludes the DriverKit SDK, which fixed the build script(s) for me.

```
$ xcodebuild -showsdks
<snip>
macOS SDKs:
	DriverKit 19.0                	-sdk driverkit.macosx19.0
	macOS 10.15                   	-sdk macosx10.15
</snip>
```

### DIFF
--- a/scripts/include.sh/build-dep-cocoapod.sh
+++ b/scripts/include.sh/build-dep-cocoapod.sh
@@ -135,7 +135,7 @@ build_git_ios()
 
 build_git_osx()
 {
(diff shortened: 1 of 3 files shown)


dinhviethoa:
Thank you!

PR #1867, opened 2020-04-14
Make mailcore buildable on Android 23+, add public key pinning, add c…

p-barabas:
Add public key pinning for iOS and Android
Buildable on Android 23+.

I have no time to cleanup the code and structure, but the pinning and client certificate features are clean.


vonox7:
https://github.com/MailCore/mailcore2/pull/1912 fixes also mailcore to be buildable again

PR #2011, opened 2024-12-01
Modify Package.swift to use source target instead of binary

jsflax:
This modifies the `Package.swift` file to use source instead of binary, fixing a number of issues users have had with adding the package to their projects. It also prevents users from having to use a stale binary.

Relevant PRs for the gitmodules:

https://github.com/dinhvh/ctemplate/pull/3
https://github.com/dinhvh/tidy-html5/pull/1
https://github.com/dinhvh/libetpan/pull/451

If this PR is accepted, there are a number of documentation changes that would also be required, since adding to SPM would be simplified.


haithngn:
Hi @jsflax,

Thank you for the PR—great work!
It’s time to move away from CocoaPods, so your help is greatly appreciated.
I’m currently reviewing this and the related PRs as well.

I’ll get back to you soon!

functionaldude:
@jsflax we started using your fork for SPM support (iOS app), however we seem to have some problems with specific mail servers where we have to use `MCOAuthType.saslPlain`
```
session.authType = MCOAuthType.saslPlain
``` 
checkAccountOperation fails with code 5

I assume the reason for this is this line in your `Package.swift`:
```
.linkedLibrary("sasl2", .when(platforms: [.macOS])),
```
because for me this implies that we cannot use SASL on iOS. 
But a workaround must exist, because previously we already used MailCore2, and we haven't encountered this issue, even when we used `MCOAuthType.saslPlain`

Any ideas how to fix this issue?

functionaldude:
FYI: cocoapods uses [this](https://github.com/jsflax/mailcore2/blob/master/cocoapods/mailcore2-ios.podspec) pre-compiled binary defined in the podspec. This binary works with `MCOAuthType.saslPlain`, but compiling it via this PR breaks that feature. Sadly i can't figure out why.
FYI: we have our own mirror for the binary [here](https://github.com/ju135/mailcore2/tree/master/cocoapods-sources) because the original page is often unreachable

jsflax:
It's odd to me that this is causing an error since libsasl is explicitly left out of the iOS podspec. Can you try removing the `.when(platforms: [.macOS])` and see if that helps? It seems to compile locally if I do that, but I'm not sure if it will fix your problem.

functionaldude:
in that case I get a compilation error
image

functionaldude:
and while that library is also missing from the podspec (probably because of this build error), it is possible, that the static library in the podspec has it included

ju135:
@jsflax the issue is that libsasl is a dependency of libetpan (see also [the script to build libetpan for iOS](https://github.com/MailCore/mailcore2/blob/7417b2e8dd7e2c028aadb72056e4d1428c0627c4/scripts/build-libetpan-ios.sh#L15))
You can also check the [prepare-cyrus-sasl.sh](https://github.com/dinhvh/libetpan/blob/master/build-mac/dependencies/prepare-cyrus-sasl.sh) inside the libetpan repo to see how the dependency is built for iOS. This needs to somehow be part of the libetpan building process.

jsflax:
@ju135 That's appreciated, I'll take a look.

ju135:
Thank you! In case it helps, I already found a way to integrate the dependency in [my fork ](https://github.com/dinhvh/libetpan/compare/master...ju135:libetpan:jsflax/master). But I'm not entirely sure if this is the proper way to do it.

jsflax:
@ju135 if the xcframework is correctly compiled for the platforms we need to support, I'm fine to use a `binaryTarget` as a stopgap to get this fixed.

If you'd like to open a PR up form your fork to mine I'll happily review.

functionaldude:
@jsflax PR is open: https://github.com/jsflax/libetpan/pull/1
no changes needed for mailcore2. This passes our internal tests for `MCOAuthType.saslPlain`.

